PR #13, opened 2024-03-15
feat: keep content block as is as possible

### LINKED ISSUE #5: Suboptimal line breaking in long sentence with function call

Enter-tainer:
should greatly fix #5


Enter-tainer:
Very hard an error prone!!

PR #23, opened 2024-03-24
feat: better handling for content block

Enter-tainer:


Enter-tainer:
idea: when processing content blocks, split contents to prefix/postfix spaces and lines. If a content block has prefix & postfix spaces, we can decide whether to use
```
[ a ] or [
  a
]
```
. Otherwise we should keep it untouched

WannesMalfait:
Will this patch turn
```typst
[
word]
```
into
```typst
[
  word
]
```

Enter-tainer:
> Will this patch turn
> ```typst
> [
> word]
> ```
> into
> ```typst
> [
>   word
> ]
> ```

maybe. but this is wrong. because this adds an extra space at the end of word. ideally we should distinguish between "inline block" and "multiline block"

PR #48, opened 2024-04-09
perf: remove flat_alt to speed up nested structure

Enter-tainer:


tests/assets/unit/code/perf-nest.typ
Enter-tainer, line 1: cc @Myriad-Dreamin FYI

PR #52, opened 2024-04-15
refactor: accept `Source` as input and make docs better

### LINKED ISSUE #51: Benefit from incremental reparsing
We can provide a formatting function receiving `typst::syntax::Source` to gain benefit from incremental reparsing. Now all mainstream typst servers can do incremental source code reparsing to reduce overhead to parse, also preserve stable `typst::syntax::Span` in AST nodes across user editions (to help cache). However, typstyle can only receive a raw string and parse from scratch inside of itself formatting procedure, which is a bit wasty.

Enter-tainer:
fix #51

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1,20 +1,68 @@
+#![doc = include_str!("../README.md")]
+#[doc(hidden)]
 pub mod attr;
+#[doc(hidden)]
 pub mod pretty;
+#[doc(hidden)]
 pub mod util;
 
+#[doc(hidden)]
 pub use attr::calculate_attributes;
+#[doc(hidden)]
 pub use pretty::PrettyPrinter;
 
-pub fn pretty_print(content: &str, width: usize) -> String {
-    let root = typst_syntax::parse(content);
-    if root.erroneous() {
-        return content.to_string();
+use typst_syntax::{Source, SyntaxNode};
+
+/// Entry point for pretty printing a typst document.
+#[derive(Debug, Clone)]
+pub struct Typstyle {
+    content: String,
+    root: SyntaxNode,
+    width: usize,
+}
+
+impl Typstyle {
+    /// Create a new Typstyle instance from a string.
+    /// # Example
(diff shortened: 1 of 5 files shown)


Enter-tainer:
image

src/lib.rs
```diff
@@ -1,20 +1,68 @@
...
+            width,
+        }
+    }
+
+    /// Create a new Typstyle instance from a [`Source`].
+    ///
+    /// This is useful when you have a [`Source`] instance and you can avoid reparsing the content.
+    pub fn with_src(src: Source, width: usize) -> Self {
+        let content = src.text().to_string();
+        let root = typst_syntax::parse(&content);
```
Enter-tainer, line 47: TODO: copilot did this wrong

src/lib.rs
```diff
@@ -1,20 +1,68 @@
...
+impl Typstyle {
+    /// Create a new Typstyle instance from a string.
+    /// # Example
+    /// ```rust
+    /// use typstyle_core::Typstyle;
+    /// let content = "#{1+1}";
+    /// let res = Typstyle::with_content(content.to_string(), 80).pretty_print();
+    /// assert_eq!(res, "#{\n  1 + 1\n}");
+    /// ```
+    pub fn with_content(content: String, width: usize) -> Self {
```
Myriad-Dreamin, line 33: A bikeshedding, The `with_xxx` functions should have signature like `fn with_xxx(self, ...) -> Self`, so the naming `with_content` and `with_src` are confusing here. It would like to be renamed:
+ `with_content` -> `new_content`
+ `with_src` -> `new_src`

Myriad-Dreamin (commented):
lgtm

src/lib.rs
Enter-tainer: https://rust-lang.github.io/api-guidelines/predictability.html#constructors-are-static-inherent-methods-c-ctor


maybe new_with_src, according to guideline

Issue #59, opened 2024-04-26
[ FEATURE ] Format #table in a human readable way


LamprosPitsillos:
https://github.com/astrale-sharp/typstfmt/discussions/170#discussion-6548338

Would it be possible to format tables in a way that makes since while reading them ?

example
```typ
#table( columns: 2,
  table.header([ Example ], [ Name ]), [ twowords ], [ flatcase ], [ TWOWORDS ], [ UPPERCASE ], [ twoWords ], [ camelCase ], [ TwoWords ], [ PascalCase ], [ two_words ], [ snake_case ], [ TWO_WORDS ], [ SCREAMING_SNAKE_CASE ], [ two_Words ], [ camel_Snake_Case ], [ Two_Words ], [ Pascal_Snake_Case ], [ two-words ], [ kebab-case ], [ TWO-WORDS ], [ SCREAMING-KEBAB-CASE ], [ Two-Words ], [ Train-Case ], [ \_two_words ], [ INTERNAL ],table.cell(colspan:2,[hello])
)
```

```typ
#table(columns:2,
table.header([Formatting],[Name]),
[twowords],[flatcase],
[TWOWORDS],[UPPERCASE],
[twoWords],[camelCase],
[TwoWords],[PascalCase],
[two_words],[snake_case],
[TWO_WORDS],[SCREAMING_SNAKE_CASE],
[two_Words],[camel_Snake_Case],
[Two_Words],[Pascal_Snake_Case],
[two-words],[kebab-case],
[TWO-WORDS],[SCREAMING-KEBAB-CASE],
[Two-Words],[Train-Case],
[\_two_words],[INTERNAL],
table.cell(colspan:2,[hello])

)


```

Enter-tainer:
tracking in #17

Enter-tainer:
The initial idea is to match patterns like `grid/table(columns: $n, $cells)`.

We may put n cells in a line if they can fit in a line. Otherwise we spread them into multiple lines.(This still doesn't sounds good. Do you have better ideas?

LamprosPitsillos:
Maybe indent the leftover ones a bit?

```
line fits 3 cells*
table(columns : $n+3 , $cells...)

~~~~

foo bar baz
       faz loo soo saz
xoo xaz raz
       laz  baz laz

```

Also maybe try to include `*span` since it will break a lot of the logic otherwise

Enter-tainer:
[...]

PR #71, opened 2024-05-20
feat: support basic table formatting

Enter-tainer:
#59


Enter-tainer:
This pr handles table which :

1. no comments
2. no spread args
3. no named args, or named args appears before all pos args
4. no table/grid.header/footer/vline/hline/cell
5. `columns` is int or array

PR #79, opened 2024-05-27
feat: support format table with function calls and header/footers

### LINKED ISSUE #78: Table formatting doesn't work on macOS M1
I use typstyle 0.1.23

image

Enter-tainer:
hopefully this can fix #78 . But i cannot tell because the author doesn't provide their full code.

### DIFF
--- a/src/pretty/table.rs
+++ b/src/pretty/table.rs
@@ -4,7 +4,18 @@ use typst_syntax::{ast::*, SyntaxKind};
 
 use crate::PrettyPrinter;
 
-use super::util::func_name;
+use super::util::{func_name, indent_func_name};
+
+const BLACK_LIST: [&str; 6] = [
+    "table.cell",
+    "table.vline",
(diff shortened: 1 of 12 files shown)


src/pretty/table.rs
```diff
@@ -62,14 +105,15 @@ impl PrettyPrinter {
...
+    indent_func_name(node) == Some("table") || indent_func_name(node) == Some("grid")
 }
 
 fn is_formatable(node: FuncCall<'_>) -> bool {
     // 1. no comments
     // 2. no spread args
     // 3. no named args or named args first then unnamed args
-    // 4. no table/grid.header/footer/vline/hline/cell
+    // 4. no table/grid.vline/hline/cell
+    // 5. if table/grid.header/footer present, they should appear before/after any unnamed args
```
Enter-tainer, line 116: although it is required, it's not checked. But maybe it's not a big deal

PR #92, opened 2024-06-12
feat: preserve newline in func call

Enter-tainer:
This is useful for fletcher


Enter-tainer:
TODO: clean up the code. It is very messy at this moment

PR #99, opened 2024-06-24
feat: add CI check for files changed

kririae:
This PR adds support for CI-related return value policy.

In `--check` mode, typstyle returns 0 (success) only if:
1. ALL files are formatted correctly, and
2. ALL files are unchanged (properly projected).

Otherwise, typstyle will return 1 if, for example:
1. Any failure occurs in formatting any file, or
2. Any file is changed (not well formatted).

This change will not affect the previous return policy when `--check` is not used.

### DIFF
--- a/src/cli.rs
+++ b/src/cli.rs
@@ -25,6 +25,9 @@ pub struct CliArguments {
     pub inplace: bool,
     #[clap(subcommand)]
     pub command: Option<Command>,
+    /// Run in 'check' mode. Exits with 0 if input is formatted correctly. Exits with 1 if formatting is required.
+    #[clap(long, default_value = "false")]
+    pub check: bool,
 }
 
 #[derive(Debug, Subcommand, Clone)]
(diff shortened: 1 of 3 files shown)


src/cli.rs
```diff
@@ -25,6 +25,9 @@ pub struct CliArguments {
     pub inplace: bool,
     #[clap(subcommand)]
     pub command: Option<Command>,
+    /// Check for file changes in CI
```
Enter-tainer, line 28:
```suggestion
    /// Run in 'check' mode. Exits with 0 if input is formatted correctly. Exits with 1 if formatting is required.
```

src/main.rs
```diff
@@ -38,15 +38,21 @@ fn is_hidden(entry: &DirEntry) -> bool {
...
+        Ok(_) => CliResults::Good,
+        Err(e) => {
+            eprintln!("{e}");
+            CliResults::Bad
+        }
+    }
 }
 
-fn execute(args: CliArguments) -> Result<()> {
+fn execute(args: CliArguments) -> Result<bool> {
```
Enter-tainer, line 54: Consider adding a new enum indicating whether it is changed or not:

```
enum FormatStatus {
  Changed, Unchanged
}
```

Enter-tainer (approved):
generally LGTM

PR #100, opened 2024-06-25
feat(build): allow disabling git info collection

tranzystorekk:
When packaging for linux, some distros (e.g. Void Linux) do not use git to obtain package source, but download source archives directly. That makes any git repository info unavailable and currently requires either adding unnecessary libgit2 dependency or patching the code out.

This change allows using `--no-default-features --features=cli,wasm` to disable the git portion if needed.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -66,8 +66,6 @@ anyhow = "1"
 vergen = { version = "8.3.1", features = [
   "build",
   "cargo",
-  "git",
-  "git2",
   "rustc",
 ] }
 
@@ -85,9 +83,10 @@ harness = false
 
 [features]
 
-default = ["cli", "wasm"]
+default = ["cli", "wasm", "git-info"]
 cli = ["clap", "once_cell", "walkdir"]
 wasm = ["wasm-bindgen"]
+git-info = ["vergen/git", "vergen/git2"]
 
 [profile.test]
 opt-level = 3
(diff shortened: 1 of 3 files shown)


Enter-tainer (approved):
thank you!

Enter-tainer:
I will publish a new version that includes this change when I leave work.

Enter-tainer:
maybe our AUR maintainer is interested in this cc @Jisu-Woniu

Enter-tainer:
Hi @tranzystorekk I just released v0.11.28 which includes this commit. The tag is already pushed to github and the release will be created when the ci is finished.

Jisu-Woniu:
Thank you for this PR. I will use that feature gate in [`typstyle`](https://aur.archlinux.org/packages/typstyle) package, once the new version is released.

Jisu-Woniu:
Is it possible to make `git-info` disabled by default? So when new default features are introduced, they will be enabled without changing build command.

Enter-tainer:
> Is it possible to make `git-info` disabled by default? So when new default features are introduced, they will be enabled without changing build command. @Enter-tainer

I'm not sure. I personally prefer enable it by default. This is useful for people who use `cargo install` or `cargo build` to install typstyle

Jisu-Woniu:
As far as I know, features are difficult to opt-out, once they are made default, so unnecessary features may be set as non-default. I would suggest a new feature like `recommended` for non-necessary features.

For now, I suggest keeping it "as is", and make the decision later.

Enter-tainer:
Make sense. It like we need to group features flags in some way like what you've mentioned. Maybe we can consider this once we have a sufficient large number of features.

PR #106, opened 2024-07-08
feat: better formatting for call chains

### LINKED ISSUE #105: Dot chain format
```typst
#{
(1,2,3).rev().rev().at(0).rev()
}
```

We might want to format it as

```typst
#{
  (1, 2, 3)
    .rev()
    .rev()
    .at(0)
    .rev()
}
```

Enter-tainer:
fix #105 cc @flaribbit


Enter-tainer:
TODO: To properly handle this we need to resolve the full chain before actually producing any formatter IR.

We also need to control when to enter the fluent mode. Ideally we should do this when ever the length of the call chain exceeds 2.

tests/snapshots/assets__check_file@cetz-manual.typ-40.snap
```diff
@@ -254,15 +258,13 @@ Marks are arrow tips that can be added to the end of path based elements that su
...
+        .mark-shapes
+        .marks {
         let name-to-mnemonic = (:)
-        for (
-          name,
-          item,
-        ) in cetz.mark-shapes.mnemonics {
+        for (name, item) in cetz
+          .mark-shapes
+          .mnemonics {
```
Enter-tainer, line 267: maybe we should only enable this for chains that contains function call? I'm not sure..

tests/snapshots/assets__check_file@cetz-manual.typ-40.snap
```diff
@@ -67,7 +67,11 @@ This is the minimal starting point:
   ```]
 Note that draw functions are imported inside the scope of the `canvas` block. This is recommended as some draw functions override Typst's function's such as `line`.
 
-#show raw.where(block: false): it => if it.text.starts-with("<") and it.text.ends-with(">") {
+#show raw.where(block: false): it => if it
+  .text
+  .starts-with("<") and it
+  .text
+  .ends-with(">") {
   set text(1.2em)
```
Enter-tainer, line 75: wtf is this??

tests/snapshots/assets__check_file@cetz-tree.typ-40.snap
```diff
@@ -142,9 +142,10 @@ input_file: tests/assets/cetz-tree.typ
...
-      children = tree.slice(1).enumerate().map((
-        (n, c),
-      ) => build-node(
+      children = tree
+        .slice(1)
+        .enumerate()
+        .map(((n, c)) => build-node(
         c,
         depth: depth + 1,
         sibling: n,
```
Enter-tainer, line 151: not indented, why

tests/snapshots/assets__check_file@tablex.typ-40.snap
```diff
@@ -3713,17 +3726,19 @@ input_file: tests/assets/tablex.typ
...
-    new_vlines = range(
-      0,
-      col_len + 1,
-    ).filter(x => vlines.filter(v => v.x == x).len() == 0).map(x => vlinex(x: x))
+    new_vlines = range(0, col_len + 1)
+      .filter(x => vlines
+          .filter(v => v.x == x)
+          .len() == 0)
+      .map(x => vlinex(x: x))
   }
```
Enter-tainer, line 3742: ugly

Enter-tainer: However, previous result is even more ugly

tests/snapshots/assets__check_file@cetz-tree.typ-40.snap
Enter-tainer: looks like we are having strang formatting for closures

Enter-tainer:
https://github.com/typst/typst/issues/4551

tests/snapshots/assets__check_file@unit-code-for-loop-line-break.typ-120.snap
```diff
@@ -13,7 +13,8 @@ input_file: tests/assets/unit/code/for-loop-line-break.typ
   } else {
     -1
   }
-  for (p, c) in (
+  for (p, c) in
+    (
```
Enter-tainer, line 17: https://github.com/typst/typst/issues/4548

tests/snapshots/assets__check_file@unit-code-for-loop-line-break.typ-80.snap
```diff
@@ -43,3 +43,19 @@ input_file: tests/assets/unit/code/for-loop-line-break.typ
...
+      .rev()
+      .rev()
+      .rev()
+      .rev()
+      .rev()
+      .rev()
+      .enumerate() {
+    11111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111
+  }
+}
```
Enter-tainer, line 61: note: this is valid in code mode, but not in markup mode

Enter-tainer:
We should maintain the current scope. If it is code mode, we can freely break call chain into multiple lines. However, it is in markup mode, We should add a paren pair to make it work

tests/snapshots/assets__check_file@cetz-tree.typ-40.snap
Enter-tainer: This is because when formatting `field access`, the last arg is not considered so it is not indented. Will resolve in next prs

PR #108, opened 2024-07-12
feat: support break for loop at `in`

Enter-tainer:


tests/snapshots/assets__check_file@unit-code-for-loop-line-break.typ-40.snap
```diff
@@ -46,16 +49,17 @@ input_file: tests/assets/unit/code/for-loop-line-break.typ
...
 #let 六面体 = {
   for (
-    pppppppppppppppppppppppppppppppppppppppppp,
-    cccccccccccccccccccccccccccccccc,
-    b,
-  ) in ((111111, 111111111, 1),) {
+      pppppppppppppppppppppppppppppppppppppppppp,
+      cccccccccccccccccccccccccccccccc,
+      b,
+    ) in ((111111, 111111111, 1),) {
```
Enter-tainer, line 55: we've add too much indent here. Ideally this should only affect things after `in`

PR #109, opened 2024-07-12
feat: put single var closure's body in a new line when closure body is multiline

Enter-tainer:
The surrounding parens are required because currently typst doesn't allow closure body presenting in new line

```
#let f = a =>
  a + 1
#f(1)
```

```
Error: expected expression
   ╭─[/main.typ:8:14]
   │
 8 │ #let f = a =>
───╯

```

### DIFF
--- a/tests/snapshots/assets__check_file@tablex.typ-40.snap
+++ b/tests/snapshots/assets__check_file@tablex.typ-40.snap
@@ -54,10 +54,9 @@ input_file: tests/assets/tablex.typ
 #let calc-mod = if typst-calc-rem-supported {
   calc.rem
 } else {
-  (
-    a,
-    b,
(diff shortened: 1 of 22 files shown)


tests/snapshots/assets__check_file@tablex.typ-40.snap
```diff
@@ -54,10 +54,9 @@ input_file: tests/assets/tablex.typ
...
   calc.rem
 } else {
-  (
-    a,
-    b,
-  ) => calc.floor(a) - calc.floor(b * calc.floor(a / b))
+  (a, b) => (
+    calc.floor(a) - calc.floor(b * calc.floor(a / b))
+  )
 }
```
Enter-tainer, line 60: This is beneficial for this case

Enter-tainer (commented):
But for simpler cases it is kind of redundant

Issue #95, opened 2024-06-14
Better handling for last arg in function call


Enter-tainer:
desired behavior

```
align(..., {
  ...
})
```
current behavior :

```
align(
  ...,
  {
    ...
  }
)
```

The idea is, if args excluding the last one can fit in a single line, then the last arg is not indented. Otherwise the bahavior is same as original

Enter-tainer:
Seems to be impossible for a generic case if we don't change the pretty printter

Enter-tainer:
https://doc.rust-lang.org/nightly/style-guide/expressions.html#combinable-expressions

QuadnucYard:
This is also the case for compact single arg, such as
```typ
#func({
  ...
})
```
which is currently forced to fold instead of fitting.

PR #113, opened 2024-07-13
fix: allow if stmt without surrounding parens

Enter-tainer:


tests/snapshots/assets__check_file@cetz-manual.typ-0.snap
```diff
@@ -29,14 +29,12 @@ input_file: tests/assets/cetz-manual.typ
...
-    )
-  }
-))
+) => if num.pos().len() < 4 {
+  numbering(
+    "1.1",
+    ..num,
+  )
+})
 #show link: set text(blue)
```
Enter-tainer, line 39: ideally, we should allow printer too choose between these 2 layout depending on the column width. It should first try current one, if it cannot fit, then switch to the previous one. But I think it is not possible in current pretty printer

Enter-tainer: #95

PR #131, opened 2024-09-16
feature: add command-line completions

### LINKED ISSUE #130: Add shell completion with clap_complete
See:

<https://docs.rs/clap_complete/latest/clap_complete/>

Will be better for packaging.

Jisu-Woniu:
Trying to add command-line completions.

Closes #130

### DIFF
--- a/src/cli.rs
+++ b/src/cli.rs
@@ -37,6 +37,13 @@ pub enum Command {
         /// The directory to format. If not provided, the current directory is used
         directory: Option<PathBuf>,
     },
+    #[cfg(feature = "completion")]
+    /// Generate shell completions for the given shell to stdout
+    Completions {
+        /// The shell to generate completions for
+        #[clap(value_enum)]
+        shell: clap_complete::Shell,
+    },
 }
 
 pub enum CliResults {
--- a/src/main.rs
+++ b/src/main.rs
@@ -111,6 +111,16 @@ fn execute(args: CliArguments) -> Result<FormatStatus> {
                     bail!("failed to format {error_count} files");
                 }
             }
+            #[cfg(feature = "completion")]
+            cli::Command::Completions { shell } => {
+                use clap::CommandFactory;
+                clap_complete::generate(
+                    *shell,
+                    &mut cli::CliArguments::command(),
+                    std::env::args().next().unwrap(),
+                    &mut std::io::stdout(),
+                );
+            }
         }
 
         return Ok(changed.into());
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -69,9 +69,9 @@ dependencies = [
 
 [[package]]
 name = "anstyle"
-version = "1.0.7"
+version = "1.0.8"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "038dfcf04a5feb68e9c60b21c9625a54c2c0616e79b72b0fd87075a056ae1d1b"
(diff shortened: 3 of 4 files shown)


Jisu-Woniu:
There is an issue when building if I try to generate completions in `build.rs`:

The `LONG_VERSION` variable here:

https://github.com/Jisu-Woniu/typstyle/blob/43ead3fb62c635c844b26a51265d1297b73b137b/src/cli.rs#L10

https://github.com/Jisu-Woniu/typstyle/blob/43ead3fb62c635c844b26a51265d1297b73b137b/src/cli.rs#L64

Is not accessible since `VERGEN_CARGO_TARGET_TRIPLE` is not available during build.

Jisu-Woniu:
The code now compiles, but `cli.rs` seems quite ugly now...

Enter-tainer:
I will review this pr when i am back from vacation (20th)

Enter-tainer:
I will look into it this weekend

src/cli.rs
```diff
@@ -37,13 +39,25 @@ pub enum Command {
...
+        /// The shell to generate completions for
+        #[clap(value_enum)]
+        shell: clap_complete::Shell,
+    },
 }
 
+#[cfg(not(build))]
+static LONG_VERSION: String = String::new();
+
+#[cfg(build)]
```
Enter-tainer, line 54: could you pls explain what these `cfg` do?

Jisu-Woniu: I included `cli.rs` in `build.rs`, but some parts of `cli.rs` is not accessible in build script due to dependency issues. So I added a `rustc-cfg` in `build.rs` to separate build script and actual building.

src/main.rs
```diff
@@ -111,6 +111,16 @@ fn execute(args: CliArguments) -> Result<FormatStatus> {
...
+            cli::Command::Completions { shell } => {
+                use clap::CommandFactory;
+                clap_complete::generate(
+                    *shell,
+                    &mut cli::CliArguments::command(),
+                    std::env::args().next().unwrap(),
+                    &mut std::io::stdout(),
+                );
+            }
         }
```
Enter-tainer, line 124: do we need this? IIUC we already generate comp files in build script

Enter-tainer (approved):
overall LGTM.

src/main.rs
Jisu-Woniu: `cargo install` won't install asset files like completion, only the binaries. So we allow users to generate the completions themselves. Also, users can use the completions on the fly like:

```bash
source <(typstyle completion bash)
```

Enter-tainer: what if we only keep the completion subcommand, and remove things build script🤔 i think this will make the code cleaner(we dont need cfgs)

Jisu-Woniu: That sounds good, maybe I can revert some changes. I can generate completion scripts from the binary.

Enter-tainer (approved):
thank you!

Jisu-Woniu:
There is a problem when I revert the changes, the generated completions are using the wrong binary name.

Jisu-Woniu:
For example, if I run `cargo run -- completions bash`, what I get is:

```bash
_target/debug/typstyle() {
    local i cur prev opts cmd
    COMPREPLY=()
    cur="${COMP_WORDS[COMP_CWORD]}"
    prev="${COMP_WORDS[COMP_CWORD-1]}"
    cmd=""
    opts=""

    for i in ${COMP_WORDS[@]}
    # ...
```

PR #133, opened 2024-09-22
fix: fix wrong path for completion scripts

Jisu-Woniu:
- **fix: fix wrong path for completion scripts**

Fix issue referenced in <https://github.com/Enter-tainer/typstyle/pull/131#issuecomment-2365456088>

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -123,7 +123,7 @@ fn execute(args: CliArguments) -> Result<FormatStatus> {
 
                 let bin_name = bin_path
                     .as_ref()
-                    .and_then(|p| p.file_name().and_then(|p| p.to_str()))
+                    .and_then(|p| p.file_stem().and_then(|p| p.to_str()))
                     .unwrap_or("typstyle");
 
                 clap_complete::generate(
--- a/src/main.rs
+++ b/src/main.rs
@@ -114,10 +114,22 @@ fn execute(args: CliArguments) -> Result<FormatStatus> {
             #[cfg(feature = "completion")]
             cli::Command::Completions { shell } => {
                 use clap::CommandFactory;
+                use std::env::{args_os, current_exe};
+
+                let bin_path = args_os()
+                    .next()
+                    .map(PathBuf::from)
+                    .or_else(|| current_exe().ok());
+
+                let bin_name = bin_path
+                    .as_ref()
+                    .and_then(|p| p.file_name().and_then(|p| p.to_str()))
+                    .unwrap_or("typstyle");
+
                 clap_complete::generate(
                     *shell,
                     &mut cli::CliArguments::command(),
-                    std::env::args().next().unwrap(),
+                    bin_name,
                     &mut std::io::stdout(),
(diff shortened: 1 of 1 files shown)


Enter-tainer:
how does tinymist solve this problem?

Jisu-Woniu:
Just my coding mistake.

Enter-tainer:
https://github.com/Myriad-Dreamin/tinymist/blob/main/crates/tinymist/src/args.rs#L49

Jisu-Woniu:
I have fixed it locally, and will squash the commits later.

Enter-tainer:
can we copy-paste tinymist's impl?

Jisu-Woniu:
Their implementation seems much too complicated for me 😔.

Jisu-Woniu:
Their implementation seems much too complicated for me 😔.

Jisu-Woniu:
They seem to be detecting shell from the `$SHELL` variable, but I consider it wrong and misleading since the variable is not related to the current running shell.

src/main.rs
```diff
@@ -114,10 +114,22 @@ fn execute(args: CliArguments) -> Result<FormatStatus> {
...
+                let bin_name = bin_path
+                    .as_ref()
+                    .and_then(|p| p.file_stem().and_then(|p| p.to_str()))
+                    .unwrap_or("typstyle");
+
                 clap_complete::generate(
                     *shell,
                     &mut cli::CliArguments::command(),
-                    std::env::args().next().unwrap(),
+                    bin_name,
```
Enter-tainer, line 132: maybe we should simply hardcode `"typstyle"` here?

Jisu-Woniu: Just in case someone symlink'ed it.

src/main.rs
```diff
@@ -114,10 +114,22 @@ fn execute(args: CliArguments) -> Result<FormatStatus> {
             #[cfg(feature = "completion")]
             cli::Command::Completions { shell } => {
                 use clap::CommandFactory;
+                use std::env::{args_os, current_exe};
+
+                let bin_path = args_os()
+                    .next()
+                    .map(PathBuf::from)
+                    .or_else(|| current_exe().ok());
```
Enter-tainer, line 122: why should we use both args and current exe?

Enter-tainer: I feel like using `"typstyle"` directly is much more easlier

Jisu-Woniu: I see, symlink is beyond the scope of our consideration. The users can recompile or replace text with `sed` if they need to change this.

Enter-tainer (approved):

Enter-tainer:
does this affect aur packaging? should i release a new version right now? i suppose it wont be a urgernt fix.

Jisu-Woniu:
Yes, but maybe you can re-tag it with `git tag --force`, since I didn't push anything to AUR due to this error.

Enter-tainer:
I've pushed v0.11.34

PR #139, opened 2024-10-08
fix: inline SPDX license in Cargo.toml

kachick:
Currently, crates.io and GitHub dependency review API are detecting the license is `non-standard`, even though the LICENSE file clarified `Apache License 2.0`

image

image

https://github.com/rust-lang/cargo/blob/2e309bd75403176790c446be32d83509eb139eca/src/cargo/util/toml/mod.rs#L1627-L1631

```
                 `license` should be used if the package license can be expressed \
                 with a standard SPDX expression.\n\
                 `license-file` should be used if the package uses a non-standard license.\n\
```


Enter-tainer (approved):
thx

kachick:
Thank you!

PR #77, opened 2024-05-26
feat: allow single item code block be put in a single line

Enter-tainer:
fix #76

PR #149, opened 2024-11-21
feat: support inline code block with simple flavor-based method

QuadnucYard:
A similar solution to #77, but simple and more acceptable for now.
The core changes are just in `convert_code_block`, together with some relevant improvements.
**Only when the code block contains only one item (without comments), and it does not follow a linebreak, we will try to put it into a single line.**

This is a context-insensitive method. We just see whether the code block contains a new line to decide whether it can be put in a single line. So, inline or not is controlled by the writer itself, and we have little control over it, except that we are really unable to squeeze them into the same line.

This respects the original structure and has some compatibility. If we enlarge the column width, it will not put any code blocks into a single line.
Unfortunately, when the source is formatted with a shorter column width, it can bring awkward results, such as
```typst
#if a < b {
  something-long-enough
} else { return }
```
But we can still manually check it, and break the block into multiple lines, which will not be formatted back.

To handle such cases, I think we can require that if the whole if-statement cannot be put into a single line (including conditions), all then-else blocks should be put into multiple lines. This requires some work about scopes.

IMO, this solution can serve as a transitional scheme until we reach an ideal one, and may make the current formatting results more acceptable to a large extent.

### DIFF
--- a/src/pretty/mod.rs
+++ b/src/pretty/mod.rs
@@ -31,6 +31,10 @@ impl PrettyPrinter {
             mode: vec![].into(),
         }
     }
+
+    fn get_fold_style<'a>(&self, node: impl AstNode<'a>) -> FoldStyle {
+        FoldStyle::from_attr(self.attr_map.get(node.to_untyped()))
+    }
 }
 
 impl PrettyPrinter {
@@ -427,7 +431,11 @@ impl PrettyPrinter {
             BoxDoc::nil(),
             (BoxDoc::text("{"), BoxDoc::text("}")),
             true,
-            FoldStyle::Never,
+            if codes.len() == 1 {
+                self.get_fold_style(code_block)
+            } else {
+                FoldStyle::Never
+            },
         );
         doc
     }
@@ -504,7 +512,7 @@ impl PrettyPrinter {
                 .group();
             res
         } else {
-            let style = FoldStyle::from_attr(self.attr_map.get(array.to_untyped()));
+            let style = self.get_fold_style(array);
             comma_seprated_items(array_items.into_iter(), style, None, None)
         }
     }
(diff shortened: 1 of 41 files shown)


src/pretty/mod.rs
```diff
@@ -427,7 +431,11 @@ impl PrettyPrinter {
             BoxDoc::nil(),
             (BoxDoc::text("{"), BoxDoc::text("}")),
             true,
-            FoldStyle::Never,
+            if codes.len() == 1 {
```
Enter-tainer, line 434: The core change in this PR: if the code block contains only one item, we will put it in a single line if there are no new lines in that item. Am I correct?

QuadnucYard: Yes. It just adds a special case for blocks with only one item.

QuadnucYard:
It looks good with new tests.
But we may need to handle code blocks with multiple empty lines further, such as
```typst
#let o = {




}

#let o = {


  true

}
```

Enter-tainer (approved):
LGTM

tests/snapshots/assets__check_file@unit-code-short-if.typ-40.snap
```diff
@@ -0,0 +1,32 @@
...
+#table(
+  fill: (x, y) => if y == 0 {
+    white.darken(15%)
+  } else { none },
+  align: (x, y) => if y == 0 {
+    center
+  } else { horizon },
+  [Hi],
+  [there],
+)
```
Enter-tainer, line 16: This doesn't look nice. This is hard to implement in wadler style pretty printer. I always want to control the display style (inline or not) of A depending on B's state. But wadler style pretty printer simply doesn't support that. I guess eventually we need to propose a new IR for pretty printer and implement it by our own.

Issue #141, opened 2024-10-14
bad paren format in math


QuadnucYard:
In the cases below, I would rather not format it than wrap the content in parentheses/braces, especially when I already wrap in the `cases`, in which the contents of `cases` should not affect formatting before.

```typst
$
  f(
    x
  ) = cases(
    "a very very very very very very very long case",
    "b very very very very very very very long case".
  )
$
$
  f(
    x
  ) = abs("aaaaaaaaaaa very very very very very very very very very very very very very very very very very long arg")
$
$
  (
    a + b
  ) = abs("aaaaaaaa very very very very very very very very very very very very very very very very very long arg")
$
$
  f[
    x
  ] = abs("aaaaaaaaaaa very very very very very very very very very very very very very very very very very long arg")
$
$
  f[
    x
  ] = cos + cos + cos + cos + cos + cos + cos + cos + cos + cos + cos + cos + cos + cos + cos + cos + cos + cos + cos + cos
$
$
  f{
    x
  } = cos + cos + cos + cos + cos + cos + cos + cos + cos + cos + cos + cos + cos + cos + cos + cos + cos + cos + cos + cos
$
```

Suggested format result (not considering binop line breaks):

```typst
$
  f(x) = cases(
    "a very very very very very very very long case",
    "b very very very very very very very long case".
  )
$
$
  f(x) = abs(
    "aaaaaaaaaaa very very very very very very very very very very very very very very very very very long arg"
  )
$
$
  (a + b) = abs(
    "aaaaaaaa very very very very very very very very very very very very very very very very very long arg"
  )
$
$
  f[x] = abs(
    "aaaaaaaaaaa very very very very very very very very very very very very very very very very very long arg"
  )
$
```

Enter-tainer:
This is indeed hard to handle. The tricky part is that whether to break or not is determined by the semantics of the equation.

For example, we may disable line breaking for `f(x)` in `f(x) = long long equation`, but we may want `f((x+y)-(x-y)) = long long equation`  to be breaked into multiple lines. Would you like proposing a rule to determine whether to break a MathDelimited into multiple lines or not?

PR #150, opened 2024-11-21
feat: do not add extra newlines to math delimited

QuadnucYard:
Since we cannot handle math very well for now, a long line is better than an ugly line.
Let the writer decide whether to break the line.

If this is PR accepted, maybe more tests are needed to ensure structure preservation.


Enter-tainer (approved):
Quick question: if original text has newlines around math delimited, will this PR affect them?

QuadnucYard:
> Quick question: if original text has newlines around math delimited, will this PR affect them?

Nothing bad. It will preserve existing line breaks.

This PR could serve as a weak fix for #141. Although it still cannot format function calls in math, it will not break intended styles.

PR #151, opened 2024-11-22
fix: context expression surrounded by redundant parens

### LINKED ISSUE #129: Context expression being parenthesized
Hello,

I noticed that (after commit `2cab26dc9d6dcd31115c65a3eec43689020ae654`), functions where the entire body is in a context keyword is parenthesized. Was this intended ?
This doesn't affect let bindings that are not functions.

I thought that, compared to how if statements are not enclosed in parentheses, this is a bit inconsistent.

Example:

<summary>Before</summary>

```typ
#let foo() = context {
  5
}
```

<summary>After</summary>

```typ
#let foo() = (
  context {
    5
  }
)
```

<summary>`if` expressions are not in parentheses</summary>

```typ
#let foo() = if true {
  5
} else {
  6
}
```

Thanks a lot and have a nice day :)

QuadnucYard:
Closes #129

Also adds more tests and fixes some typos.

### DIFF
--- a/src/pretty/mod.rs
+++ b/src/pretty/mod.rs
@@ -9,7 +9,7 @@ use pretty::BoxDoc;
 use typst_syntax::{ast, ast::*, SyntaxKind, SyntaxNode};
 
 use crate::attr::Attributes;
-use crate::util::{comma_seprated_items, pretty_items, FoldStyle};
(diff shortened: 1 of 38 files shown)


Enter-tainer (approved):
thx for adding new tests!

PR #153, opened 2024-11-22
feat: format comments and commented blocks in code block & markup

QuadnucYard:
The first step in formatting commented code.

For block comments, I reference the default style of rustfmt. There are the following cases:
- Only one line: Do nothing.
- Every followed line starts with a bullet `*`: All aligned with the open marker `/*`.
- Otherwise, they are indented or dedented together to make the line with the least indentation aligned with the open `/*`.


tests/snapshots/assets__check_file@cetz-tree.typ-0.snap
```diff
@@ -64,10 +65,18 @@ input_file: tests/assets/cetz-tree.typ
...
+        source-name + "." + direction,
+        target-name + "." + opposite-dir.at(direction),
+      )
 
-      draw.line(a, b)
+      draw.line(
+        a,
+        b,
+      )
       /*
```
Enter-tainer, line 80: looks like this pr indeed fix the no format bug.

QuadnucYard:
I estimate that there is still a lot of work to implement comment-related formatting fully and correctly.
If you prefer, we can merge this first and release a new version.

PR #154, opened 2024-11-23
feat: format list-like objects with comments

QuadnucYard:
This PR rewrites the pretty conversion of list-like objects, including arrays, dicts, destructions, and parameters (and closure definitions).
It supports comments within these objects. Each item can be attached with comments at the front or back.

It also fixes:
- An array with one item is prevented from being split into multiple lines when the column is narrow.
- Patterns in destructuring may be lost after formatting.

Except that, the styles remain unchanged (if formattable before).

Some examples:

```typst
#(
   555,  // 67654
          999  /* 65543*/ ,
    123,
 /* 65543*/
    // 454456
  /* 1 */     546, /* 65543*/
       777 /* 65543*/,
      /* 65543*/
  789, 1445, 654,

  /* 5 */     666, 777, /* 659 */

)

#(1  ,   2  ,   /* */ 3  )
#(   1 , 2, /* */     3,)
#(   1 ,2 /* */ ,  3 , )
#(1,2,  3, /* */   )
#(1,)
#( /* */1,)
#( 1/* */,)
#( 1,/* */)

#( /* */..() , )
#( ../* */()  ,)
#( ..  () /* */,)
#( ..()  ,/* */  )
```
may be formatted to
```typst
#(
  555, // 67654
  999, /* 65543*/
  123,
  /* 65543*/
  // 454456
  /* 1 */ 546, /* 65543*/
  777, /* 65543*/
  /* 65543*/
  789,
  1445,
  654,
  /* 5 */ 666,
  777, /* 659 */
)

#(1, 2, /* */ 3)
#(1, 2, /* */ 3)
#(1, 2 /* */, 3)
#(1, 2, 3 /* */)
#(1,)
#(/* */ 1,)
#(1 /* */,)
#(1 /* */,)

#(/* */ ..(),)
#(../* */(),)
#(..() /* */,)
#(..() /* */,)
```
and
```typst
#(/* 1 */ // first comment
   a:    /* 2 */
/* 3 */   1/* 4 */     ,   // second comment
    b   /* 5 */    :/* 6 */ 2,/* 7 */c   : 3/* 8 */ ,
   /* 9 */ d/* 10 */
/* 11 */    :/* 12 */

/* 13 */
   /* 14 */ 5/* 15 */
/* 16 */)
```
to 
```typst
#(
  /* 1 */
  // first comment
  a: /* 2 */ /* 3 */ 1, /* 4 */ // second comment
  b /* 5 */: /* 6 */ 2,
  /* 7 */ c: 3, /* 8 */
  /* 9 */ d /* 10 */ /* 11 */: /* 12 */ /* 13 */ /* 14 */ 5, /* 15 */
  /* 16 */
)
```

It also fixes a bug with destruction. Previously,
```typst
#let (a, b: (.., d), ..c) = (a:1, b:(c: 4, d: 5))
```
will be formatted to
```typst
#let (a, b: b, ..c) = (a: 1, b: (c: 4, d: 5))
```
Now the pattern is handled correctly.


tests/snapshots/assets__check_file@packages-codly.typ-0.snap
```diff
@@ -428,10 +430,14 @@ snapshot_kind: text
...
       highlights
-        .sorted(key: x => (
+        .sorted(key: (
+            x,
+          ) => (
             x.line
           ))
-        .map(x => {
+        .map((
+        x,
```
QuadnucYard, line 439: bad indent in chain

QuadnucYard:
A big challenge is func-call and chain-access. It can be divided into two cases, not considering comments:
```typst
#{
  // If we can put everything except the arguments of the last one in one line and the last one is *heavy enough*.
  a.b().c().d(
    e(..)
  )
  
  // Otherwise, chain them.
  a
    .b(..)
    .c(..)
    .d(e(..))
}
```

We don't want to see an item in the middle split into multiple lines while there exists `.a().b()`.

tests/snapshots/assets__check_file@tablex.typ-40.snap
```diff
@@ -1041,11 +1043,9 @@ snapshot_kind: text
 
 // Fetches an entire column of cells (all positions with the given x).
 #let grid-get-column(grid, x) = {
-  range(grid-count-rows(grid)).map(y => grid-at(
-    grid,
-    x,
+  range(grid-count-rows(grid)).map((
```
QuadnucYard, line 1046: We may need to better handle this case.

QuadnucYard: And we can add an option to put the arguments on the same line, e.g. 
```typst
grid-at(
  grid, x
)
```

QuadnucYard:
Some insights about comment handling.

We can abstract a line as `before item, after`, where `before` and `after` are comments.
When we see:
- a comment, we add it to `free_comments` for future use.
- a list item, if we have any free comments, join them together.
- a comma, we attach all free comments to the last item we saw (if any).
- a line break, if there is an item in the same line, attach free comments to the last one.

If the items are in a single line, we put the attached block comment before the comma. Otherwise after the comma. So the comma positions may change.

tests/snapshots/assets__check_file@tablex.typ-40.snap
Enter-tainer: I initially want to do this but I quickly figure out rustfmt never format code like this. So I decide to follow rustfmt's design

src/pretty/entry.rs
```diff
@@ -0,0 +1,117 @@
+use pretty::BoxDoc;
+use typst_syntax::{ast::*, SyntaxKind};
+
+use super::{comment::comment, util::is_comment_node, PrettyPrinter};
+
+// The comments can only appear in the middle.
+pub struct EntryWorker<'a> {
```
Enter-tainer, line 7: what is an entry? I suggest we should choose a better name

src/pretty/entry.rs
```diff
@@ -0,0 +1,117 @@
...
+                continue;
+            }
+            self.leading = false;
+        }
+
+        self.doc.group()
+    }
+
+    fn push(&mut self, doc: BoxDoc<'a, ()>) {
+        let a = std::mem::replace(&mut self.doc, BoxDoc::nil());
```
Enter-tainer, line 106: this looks weird... What's this for?

src/pretty/list.rs
```diff
@@ -0,0 +1,276 @@
...
+use typst_syntax::{ast::*, SyntaxKind, SyntaxNode};
+
+use crate::util::FoldStyle;
+
+use super::{comment::comment, util::is_comment_node, PrettyPrinter};
+
+pub enum CommentedItem<'a> {
+    Comment(BoxDoc<'a, ()>),
+    Commented(BoxDoc<'a, ()>, BoxDoc<'a, ()>),
+}
```
Enter-tainer, line 11: Please consider use a more descriptive name for enum variants. You may also want to use enum structs to assign names for each field. This should be better than unnamed tuples.

src/pretty/list.rs
```diff
@@ -0,0 +1,276 @@
...
+        &mut self,
+        list_node: &'a SyntaxNode,
+        item_converter: impl Fn(T) -> BoxDoc<'a, ()>,
+    ) {
+        // Each item can be attached with a comment at the front and back.
+        // Can break line after the front attachment.
+        // If the back attachment appears before the comma, the comma is move to its front.
+        // Only one can attach after.
+
+        self.fold_style = self.printer.get_fold_style_ut(list_node);
```
Enter-tainer, line 129: `ut` -> `untyped` pls

src/pretty/list.rs
```diff
@@ -0,0 +1,276 @@
...
+                self.try_attach();
+            } else if node.kind() == SyntaxKind::Space {
+                let newline_cnt = node.text().chars().filter(|c| *c == '\n').count();
+                if newline_cnt > 0 {
+                    self.attach_or_detach();
+                    self.can_attach = false;
+                }
+            }
+        }
+        self.attach_or_detach();
```
Enter-tainer, line 162: This is not a good name. It sounds like it would return a bool and it takes `&self`. But instead it returns nothing and mutates `self`

src/pretty/list.rs
```diff
@@ -0,0 +1,276 @@
...
+                        open.append(inner).append(close)
+                    }
+                }
+            };
+            multi.clone().flat_alt(flat).group()
+        }
+    }
+}
+
+fn is_exact_one_and<T>(mut iterator: impl Iterator<Item = T>, f: impl FnOnce(&T) -> bool) -> bool {
```
Enter-tainer, line 272:
```suggestion
fn is_only_one_and<T>(mut iterator: impl Iterator<Item = T>, f: impl FnOnce(&T) -> bool) -> bool {
```

src/pretty/mod.rs
```diff
@@ -636,48 +586,26 @@ impl PrettyPrinter {
...
-                && !matches!(
-                    closure.params().children().next().unwrap(),
-                    Param::Pos(Pattern::Destructuring(_))
-                )
-            {
-                doc = params[0].clone();
-            } else {
-                doc = arg_list
-            }
-            doc = doc
```
Enter-tainer, line 667: pls bring this back. This makes `it => body` possible

src/pretty/entry.rs
Enter-tainer: I havn't look into this in detail, but I guess we should simply use `Option<BoxDoc>` here.

src/pretty/entry.rs
QuadnucYard: > what is an entry? I suggest we should choose a better name

I mean an entry of dict. This file is mainly designed for DictItem, but also used in `DestructuringItem` and `Param`. Name it `item`? Do you have a better idea?

In the beginning I want to name it just `dict.rs`, but this also cannot describe the functionality of the module well.

src/pretty/entry.rs
QuadnucYard: I want to simplify `self.doc = self.doc.append(...);`, but it has issues with move.
Maybe we should not be lazy here.

QuadnucYard:
> BTW, is dot chain formatting still a problem? I haven't examine snapshots yet. I can help take a look if the problem still present.

I haven't touched it yet, so the chain formatting is expected to remain unchanged.
I will work on it in the next PR.

tests/snapshots/assets__check_file@unit-code-for-loop-line-break.typ-80.snap
```diff
@@ -28,7 +28,9 @@ snapshot_kind: text
     pppppppppppppppppppppppppppppppppppppppppp,
     cccccccccccccccccccccccccccccccc,
     b,
-  ) in ((111111, 111111111, 1),) {
+  ) in (
+    (111111, 111111111, 1),
+  ) {
     "111111111111111111111111111111111"
   }
```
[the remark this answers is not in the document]
QuadnucYard: It just has a line break before formatting!

Enter-tainer (approved):
LGTM

Enter-tainer:
> It also fixes a bug that an array with one item is not broke into multiple lines when the column is narrow.

> For closure parameters, I do not make "one parameter" a special case (although it easy to implement), to align with the format behavior of multiple parameters. So this is a breaking change. Since the structure of a closure has not been handled very well yet, this change may cause dissatisfaction.

BTW please consider edit the PR description to reflect latest changes. thanks!

Enter-tainer (approved):
Thanks!

PR #160, opened 2024-11-26
test: add test case for dot chain call

Enter-tainer:
weird things:

1. all dot chain params are not breaked into multiple lines, except for the last one
2. if the last item is breaked into multiple lines, the indent is not correct


Enter-tainer:
For 1, it is because the last function call is out of FieldAccess so we cannot resolve them

For 2, I guess it's a feature, otherwise it would be ugly. 
```
#{
  a
    .b(
      1,
      2,
    )
    .c(
    3,
    4,
  )
}
```

PR #161, opened 2024-11-27
feat: format most commented elements

QuadnucYard:
This PR makes the most elements formattable. The only untouched expressions are `Parenthesized`, `FieldAccess`, and `FuncCall`, which have special treatment. Other elements are all marked formattable if not in math. No style change.
Fixes a bug with destructuring introduced in #154. Destruct and params with line comments should not fold into one line.

We treat elements focused on in this PR as **flow of items**. They are joined by spaces (can be optional) or blocks, and no line break is allowed within.

We use a trick. When formatting these elements, we need not care about the syntax, but simply consider their types and make corresponding countermeasures. In some cases where Expr and Pattern coexist, we need to record the order.

The implementations are subject to change for simplicity and expressiveness.


src/pretty/flow_like.rs
Enter-tainer, line 1: How about rename this file to `flow.rs`?

src/pretty/flow_like.rs
```diff
@@ -0,0 +1,114 @@
...
+
+use super::{util::is_comment_node, ArenaDoc, PrettyPrinter};
+
+pub struct FlowItem<'a>(Option<FlowItemRepr<'a>>);
+
+struct FlowItemRepr<'a> {
+    doc: ArenaDoc<'a>,
+    /// Whether a space can be added before the doc. Not enforcement.
+    space_before: bool,
+    ///  Whether a space can be added after the doc.
```
Enter-tainer, line 12:
```suggestion
    /// Whether a space can be added after the doc.
```

Enter-tainer: Is it enforced or not?

src/pretty/flow_like.rs
```diff
@@ -0,0 +1,114 @@
...
+        Self::doc(doc, true)
+    }
+
+    pub fn tight_doc(doc: ArenaDoc<'a>, space_after: bool) -> Self {
+        Self(Some(FlowItemRepr {
+            doc,
+            space_before: false,
+            space_after,
+        }))
+    }
```
Enter-tainer, line 35: These constructors are confusing..

Enter-tainer: Consider adding a constructor `new` that takes doc, space_before, space_after. And implement rest of them using `new`

src/pretty/flow_like.rs
```diff
@@ -0,0 +1,114 @@
...
+impl<'a> FlowItem<'a> {
+    pub fn doc(doc: ArenaDoc<'a>, space_after: bool) -> Self {
+        Self(Some(FlowItemRepr {
+            doc,
+            space_before: true,
+            space_after,
+        }))
+    }
+
+    pub fn doc_spc(doc: ArenaDoc<'a>) -> Self {
```
Enter-tainer, line 25:
```suggestion
    pub fn doc_space(doc: ArenaDoc<'a>) -> Self {
```

src/pretty/flow_like.rs
```diff
@@ -0,0 +1,114 @@
...
+        }))
+    }
+
+    pub fn none() -> Self {
+        Self(None)
+    }
+}
+
+pub struct FlowStylist<'a> {
+    arena: &'a Arena<'a>,
```
Enter-tainer, line 43: I think we can get arena using printer?

src/pretty/flow_like.rs
```diff
@@ -0,0 +1,114 @@
...
+        mut producer: impl FnMut(&'a SyntaxNode) -> FlowItem<'a>,
+    ) -> ArenaDoc<'a> {
+        let mut flow = FlowStylist::new(self);
+        for child in node.children() {
+            if child.kind().is_keyword() {
+                flow.push_doc(self.arena.text(child.text().as_str()), true, true);
+            } else if is_comment_node(child) {
+                flow.push_comment(child);
+            } else {
+                let item = producer(child);
```
Enter-tainer, line 95: `producer` is a clever invention

src/pretty/flow_like.rs
Enter-tainer: pls also document the difference between all of them. 🙇‍

src/pretty/func_call.rs
```diff
@@ -83,7 +83,7 @@ impl<'a> ParenthesizedFuncCallArg<'a> {
 impl<'a> PrettyPrinter<'a> {
     pub(super) fn convert_func_call(&'a self, func_call: FuncCall<'a>) -> ArenaDoc<'a> {
         let mut doc = self.convert_expr(func_call.callee());
-        if let Some(res) = self.check_disabled(func_call.args().to_untyped()) {
+        if let Some(res) = self.check_comment_unable(func_call.args().to_untyped()) {
```
Enter-tainer, line 86: `check_comment_unable` is a super weird name.

src/pretty/list.rs
```diff
@@ -257,9 +289,14 @@ impl<'a> ListStylist<'a> {
...
-            FoldStyle::Never => multi,
-            FoldStyle::Fit => multi.clone().flat_alt(flat()).group(),
+        let fold_style = if self.has_line_comment {
+            FoldStyle::Never
+        } else {
+            self.fold_style
+        };
+        match fold_style {
+            FoldStyle::Never => multi(),
+            FoldStyle::Fit => multi().flat_alt(flat()).group(),
```
Enter-tainer, line 299: btw, flat_alt is dangerous. It can be O(2^n) if used recursively. Previously i've removed it in most places. see https://enter-tainer.github.io/typstyle/changelog/#v01112---2024-04-09

tests/assets/unit/comment/comment-in-import.typ
```diff
@@ -0,0 +1,17 @@
...
+#import"@preview/fletcher:0.5.2"as fletcher:(
+  node,edge
+
+  )
+
+#import/* 0 */"test.typ"/* 1 */:/* 2 *//* 3 */*
+#import/* 0 */"test.typ"/* 1 */:/* 2 */a./* 3 */b/* 4 */.c. /* 5 */d, /* 6 */  eee/* 7 */as /* 8 *//* 9 */   fff
+#import/* 0 */"@preview/fletcher:0.4.0"/* 1 */as/* 2 */fletcher/* 3 */:/* 4 */ node  /* 5 */  ,   /* 6 */edge
+#import"@preview/fletcher:0.5.2"as fletcher:/* 0 */(/* 1 */node/* 2 */,/* 3 */edge/* 4 *//* 5 */)
+#import"@preview/fletcher:0.5.2"as fletcher:(/* 0 */
```
Enter-tainer, line 14: is it valid?😰

Enter-tainer (approved):
Overall LGTM. The idea of flow is fantastic.

Could you please also edit the changelog and the limitation doc to reflect latest changes? For example, https://enter-tainer.github.io/typstyle/limitations/#when-there-are-block-comment-child-gives-up-formatting-the-whole-node is no longer valid

tests/assets/unit/comment/comment-in-import.typ
QuadnucYard: it looks weird, but really valid!

src/pretty/flow_like.rs
QuadnucYard: I did not make it clear enough. Actually, a space is added only when the item before and the item after both allow this.

src/pretty/func_call.rs
QuadnucYard: I struggled with naming here. I want to convey that the node is set as unformatteable and unable to format due to insufficient capability.

src/pretty/list.rs
QuadnucYard: TBH, this function derives from your `pretty_items`, with the general logic kept. I came to realize that `flat_alt` is not necessary on many occasions not long ago.

QuadnucYard: Consider structures that behave differently in the 'break' style, and we can't decide whether to break the line beforehand: with a trailing comma (such as array), optional parens.
Can we still avoid `flat_alt` in these cases?

QuadnucYard: Maybe we can add an `auto` case with flexible `line()`s, and make the existing `flat_alt` cover fewer docs.

src/pretty/flow_like.rs
Enter-tainer: Make sense. It depends on you to decide whether to rename it or not.

src/pretty/func_call.rs
Enter-tainer: How about `check_unformattable`?

src/pretty/list.rs
Enter-tainer: > this function derives from your pretty_items

Correct. But since v0.11.12 pretty items is not used very frequently. It is mostly replaced by another function with similar capability. I cannot remember its name though but I think you should have seen it before.

> Can we still avoid flat_alt in these cases?

The idea is that we *can* use flat_alt, but only in leaf nodes.

src/pretty/flow_like.rs
QuadnucYard: BTW, do you think it is a bad idea to just return Option((Doc, bool, bool))?

Enter-tainer: pls dont, it's worse, because we dont know what these bools mean now

Enter-tainer:
Thanks!

PR #165, opened 2024-11-30
feat: format commented parenthesized expr/patterns

QuadnucYard:
This PR includes:
- feature: parenthesized expressions can be formatted just like arrays (the rules are similar to array). Nested parentheses will not be removed if there exists any comment as a direct child.
- feature: remove unnecessary parentheses if the inner expression is literal, array, dict, destructuring, block, or pattern. For safety, parens around idents are kept.
- change: destructuring and params with comments are no longer forced to fold into one line.
- feature: format closures with comments, which was left out in the previous PR.


tests/snapshots/assets__check_file@tablex.typ-120.snap
```diff
@@ -743,13 +749,19 @@ snapshot_kind: text
 
 // Returns 'true' if the cell at (x, y)
 // exists in the grid.
-#let grid-has-pos(grid, x, y) = (grid-index-at(x, y, grid: grid) < grid.items.len())
+#let grid-has-pos(grid, x, y) = (
+  grid-index-at(x, y, grid: grid) < grid.items.len()
```
QuadnucYard, line 753: This respects the multiline flavor

tests/snapshots/assets__check_file@tablex.typ-120.snap
```diff
@@ -1919,11 +1933,13 @@ snapshot_kind: text
...
-                or v.end >= cell.y + cell.rowspan  // ends at or after this cell
-            )
+          v.end in (auto, none) // always goes towards the bottom
+          or v.end >= cell.y + cell.rowspan // ends at or after this cell
+        )
 
-        (at_left_or_right and vline_hasnt_already_ended)
+        (
+          at_left_or_right and vline_hasnt_already_ended
+        )
```
QuadnucYard, line 1942: There is a line break before `and`:
```
            (at_left_or_right
                and vline_hasnt_already_ended)
```

Enter-tainer:
> feature: parenthesized expressions can be formatted just like arrays

Could you elaborate?

Enter-tainer (approved):
I went through all the snapshot changes and I didn't see anything super weird. So, LGTM

QuadnucYard:
> > feature: parenthesized expressions can be formatted just like arrays
> 
> Could you elaborate?

A parenthesized expression differs from an array only in that it only has one item, the delimiter is optional if single line, and no trailing comma.

PR #167, opened 2024-12-01
feat: format simple commented func-calls

QuadnucYard:
I extended my `ListStylist` to enable it to handle line breaks, and reimplemented formatting code blocks with it.
With this PR, function calls with comments are made formattable. I have not touched "as-is" yet.

There is a minor style change:
In code mode, the multiline flavor of a node now only depends on line breaks in its direct children, as opposed to its descendants.

After this, the only unhandled cases of comments are supposed to be structured tables and chained accesses.


tests/snapshots/assets__check_file@typstfmt-160-fletcher.typ-40.snap
```diff
@@ -2,26 +2,59 @@
...
+    node(
+      (+1, 0),
+      [variable],
+      radius: 3em,
+    ), // test
+    node(
+      (+1.25, .75),
+      [const],
+      radius: 3em,
+    ), // this is the second test
```
Enter-tainer, line 21: previously we dont format this because of the presence of comments?

src/attr.rs
```diff
@@ -17,6 +17,7 @@ pub struct Attributes {
     pub commented: bool,
     /// If (a) the first space child contains a newline, (b) one of the children is a multiline
     pub multiline: bool,
+    pub multiline_self: bool,
```
Enter-tainer, line 20: i guess this correspond to case (a) ?

QuadnucYard: yes, as mentioned in the PR description

Enter-tainer: Maybe we should have a more descriptive name(`multiline => is_descent_multiline`, `multiline_self` => `is_direct_children_multiline` for example) and add doc string.

Enter-tainer (approved):
I go through all the snapshots and they look amazing. Overall LGTM!

src/pretty/plain.rs
```diff
@@ -0,0 +1,110 @@
...
+enum PlainItem<'a> {
+    Item(ArenaDoc<'a>),
+    Comma,
+    Linebreak(usize),
+    LineComment(ArenaDoc<'a>),
+    BlockComment(ArenaDoc<'a>),
+}
+
+/// A stylist that keeps the original structure.
+pub struct PlainStylist<'a> {
```
Enter-tainer, line 16: Just wondering: we are having multiple stylist now. Do they have similar interface? If so, we can add a trait for Stylist

tests/snapshots/assets__check_file@unit-grid-basic-cmt.typ-120.snap
```diff
@@ -4,13 +4,19 @@ expression: doc_string
...
+#table(
+  columns: 2, [Name], [Age], [Hannes], [30], /* 3 */
+)
+#table(
+  columns: 2, [Name], [Age], /* 4 */ [Hannes], [40], /* 5 */
+)
+#table(columns: 2, [Name], [Age], /* 6 */ [Hannes], /* 7 */ [50],)
+#table(
+  columns: 2, [Name], /* 8 */ [Age], [Hannes], [60], // wwww
 )
```
Enter-tainer, line 20: they look very different after the change. Is this expected?

QuadnucYard: I modified this case. Everything meets expectations.

src/pretty/plain.rs
QuadnucYard: I don't have an idea of how to better abstract and reuse these stylists at the moment.
The `FlowStylist` is the simplest one that can only handle spacings, and can be used elsewhere.
The `PlainStylist` is like `ListStylist` (accepts an iterable and creates a doc), but doesn't provide flexible fitting. I think it's not time for further refactoring yet.

Enter-tainer (approved):
Thanks!

PR #171, opened 2024-12-05
feat: format binary chains and chains with comments

QuadnucYard:
Closes #114

## Features

- (feature) Format binary expressions (except assignment) as chains. Parentheses are added if necessary.
- (feature) Support formatting chains with comments. This is the last piece of formatting with comments.
- (feature) Break dot chains in markup with parentheses, if the chain contains at least two dots and one function calls.

## Limitations

Just as args in funcall, we could not make the last item special. For example, We could not achieve this:
```typ
#a.b.c(
  ..
)
#(
  a + b + (
    ..
  )
)
```

PR #173, opened 2024-12-07
feat: better args formatting

### LINKED ISSUE #143: Strange formatting of long show-set rule
```typ
#show heading.where(level: 2): set heading(numbering: (..n) => [#n.pos().at(1).])
```

with `typstyle -c 80`:

```typ
#show heading.where(level: 2): set heading(numbering: (
  ..n,
) => [#n.pos().at(1).])
```

I think it's much more readable to prefer first breaking by 1 parameter per line:

```typ
#show heading.where(level: 2): set heading(
  numbering: (..n) => [#n.pos().at(1).],
)
```

QuadnucYard:
This PR includes:

- A simple optimization. For verbatim leaf text conversions, we do not need to go through the heavy logic in `trivia`. For verbatim inner text conversions, we only need to borrow the source text.
- A breaking style change. We no longer force every single arg of simple expressions to fold. This may bring more flexibility. With the previously implemented chain formatting (#171), func-calls will not become so weird.
  In the current implementation, only when the only arg is positional or spread, and it is not `FuncCall`, `FieldAccess`, `Unary`, or `Binary`, will it be forced to fold. Otherwise, it will fit regardless of flavor.
  According to the snapshots, this has little effect on width -80 and -120.
  Fixes #143.

Before we truly implement compact formatting as in Rust (we all agree that `pretty` cannot achieve this efficiently), we cannot make the result in every context satisfactory.


src/pretty/mod.rs
```diff
@@ -163,7 +164,25 @@ impl<'a> PrettyPrinter<'a> {
...
+    }
+
+    /// For leaf only.
+    fn convert_verbatim(&'a self, node: impl AstNode<'a>) -> ArenaDoc<'a> {
+        self.convert_verbatim_untyped(node.to_untyped())
+    }
+
+    /// For leaf only.
+    fn convert_verbatim_untyped(&'a self, node: &'a SyntaxNode) -> ArenaDoc<'a> {
+        self.arena.text(node.text().as_str())
```
Enter-tainer, line 185: maybe add a debug assert here to prevent incorrect usage

src/pretty/mod.rs
```diff
@@ -286,6 +289,7 @@ impl<'a> PrettyPrinter<'a> {
     }
 
     fn convert_raw(&'a self, raw: Raw<'a>) -> ArenaDoc<'a> {
+        println!("raw ");
```
Enter-tainer, line 292: oops

tests/snapshots/assets__check_file@cetz-manual.typ-120.snap
```diff
@@ -28,9 +28,11 @@ snapshot_kind: text
...
 #set par(justify: true)
-#set heading(numbering: (..num) => if num.pos().len() < 4 {
-  numbering("1.1", ..num)
-})
+#set heading(
+  numbering: (..num) => if num.pos().len() < 4 {
+    numbering("1.1", ..num)
+  },
+)
 #show link: set text(blue)
```
Enter-tainer, line 36: cool

tests/snapshots/assets__check_file@cetz-tree.typ-40.snap
```diff
@@ -107,10 +109,12 @@ snapshot_kind: text
...
-        panic("Unsupported content type "
-          + type(content)
-          + "! "
-          + "Provide your own `draw-node` implementation.")
+        panic(
+          "Unsupported content type "
+            + type(content)
+            + "! "
+            + "Provide your own `draw-node` implementation.",
+        )
```
Enter-tainer, line 117: this looks nice. But I guess the trailing comma here is unneeded? I'm not quite sure at this moment

QuadnucYard: needless, but the args are multiline, thus a trailing comma is added

src/pretty/mod.rs
QuadnucYard: actually, the previously frequently used `trivia` is also based on this assertion.
What I didn't have time to point out in the description is that, compared with `trivia`, this `verbatim` has fewer checks for inline items, and may be more performant.

QuadnucYard: unfortunately, we could not check whether the node is leaf.

src/pretty/func_call.rs
```diff
@@ -55,48 +56,48 @@ impl<'a> PrettyPrinter<'a> {
...
+                    | Expr::Conditional(_)
+                    | Expr::While(_)
+                    | Expr::For(_)
+                    | Expr::Contextual(_)
+                    | Expr::Raw(_)
+            )
+        }) {
+            fold_style = FoldStyle::Always;
+        }
+
```
Enter-tainer: this makes a lot of sense to me.

tests/snapshots/assets__check_file@packages-codly.typ-120.snap
```diff
@@ -943,13 +943,7 @@ snapshot_kind: text
...
 
   // Get the height of an individual line.
-  let line-height = (
-    measure(
-      raw.line(1, 1, "Hello, World!", [Hello, World!]),
-    ).height
-      + padding.top
-      + padding.bottom
-  )
+  let line-height = measure(raw.line(1, 1, "Hello, World!", [Hello, World!])).height + padding.top + padding.bottom
```
Enter-tainer, line 947: This becomes a single line. (which is expected)

Enter-tainer (approved):
LGTM!

PR #174, opened 2024-12-07
feat: always add end of file newline

Enter-tainer:
Always add end of file `\n`


Myriad-Dreamin:
how does rustfmt do it? It looks like you did yet another flavor checking about the line feed before eof.

Enter-tainer:
> how does rustfmt do it? It looks like you did yet another flavor checking about the line feed before eof.

it's old behavior. in this pr, eof newline is always added

Enter-tainer:
It looks like rustfmt always add trailing newline

Myriad-Dreamin (approved):

PR #175, opened 2024-12-08
fix: performance regression from text retrieval

QuadnucYard:
The benchmark shows that the performance for documents with massive unformattable texts declined significantly in #173.
This PR reverts the method that retrieves text from the source.


Enter-tainer (approved):
thx. does this give better benchmark result?

Issue #82, opened 2024-05-28
Format of code block inside content block


8LWXpg:
Indent in consecutive code block after `[` is doubled

```typst
#[#while list.len() > 0 {
    list.pop()
  }\ ]
```

Desired output

```typst
#[#while list.len() > 0 {
  list.pop()
}\ ]
```

or

```typst
#[
  #while list.len() > 0 {
    list.pop()
  }\ 
]
```

Enter-tainer:
related #21

Enter-tainer:
https://github.com/Enter-tainer/typstyle/pull/23#issuecomment-2016806242

Enter-tainer:
Note: in typst, `[something]` and `[ something ]` is different: the second one has extra spaces

QuadnucYard:
It is still tricky to decide whether we can add a linebreak in the front of the content block based on context.
For example,
```typ
// empty line here

#[foo
bar ]
```
It involves risks to format as
```typ
// empty line here

#[
  foo
  bar
]
```
although the output is consistent with before.

If we move the content block in this case to another context, we may forget that it is the original context that makes adding a space before `foo` safe.

PR #178, opened 2024-12-09
feat: turn side spaces into linebreaks in multiline markup

QuadnucYard:
This PR includes:

- (performance improvement) Optimize attr computation by reducing HashMap access. The time consumption in this stage is expected to reduce by about 30%.
- (style change) Change the flavor decision for equations to make the behavior more consistent. Now it will break the line beside `$` as long as the equation contains a linebreak. Previously, it simply considered whether the first space contained any linebreak.
- (feature) In a multiline markup block, the leading or trailing space will turn to linebreak, but not vice-versa. A single-line content block with side spaces may carry ambiguity, so we do not fold a multiline content block into a single line. It is a safe solution to #82.
- (fix) The leading space (not leading linebreaks) of the document is now stripped. This space makes no difference to the output.

## Details of content line-breaking

- The space around list/enum/term items is stripped.
- In a multiline content block, the leading and trailing spaces (if both exist) turn to hard lines.
- In a single-line content block, the leading and trailing spaces (if both exist) turn to soft lines.
- Otherwise, the leading/trailing space is kept if the content block is single-line, and turns to a hard line if the content block is multiline.


QuadnucYard:
Oh it is not ok for convergence to keep the space.

For instance,
```typ
#[ #read("large-cell.typ", encoding: "utf8") ]
```
can be formatted to
```typ
#[ #read(
  "large-cell.typ",
  encoding: "utf8",
) ]
```
if we keep the space as it was. But then the content block is considered as multiline, and the space is turned to linebreak:
```typ
#[
  #read(
    "large-cell.typ",
    encoding: "utf8",
  )
]
```

However, if we make the space `line()`, the following case will look bad:
```typ
#[, ]
```
to
```typ
#[,
]
```

QuadnucYard:
The space to the left of `:` in term-items is stripped. This makes for consistency and avoids linebreaks.

tests/snapshots/assets__check_file@cetz-manual.typ-40.snap
```diff
@@ -336,7 +336,8 @@ line(..c)
     "array",
     "dictionary",
   ),
-  [This option sets the mark to draw when using the `mark` draw function, or applies styling to both mark ends of path based elements. The mark's name or shorthand can be given, multiple marks can be drawn by passing an array of names or shorthands. When `none` no marks will be drawn. A style `dictionary` can be given instead of a `string` to override styling for that particular mark, just make sure to still give the mark name using the `symbol` key otherwise nothing will be drawn!. ],
+  [This option sets the mark to draw when using the `mark` draw function, or applies styling to both mark ends of path based elements. The mark's name or shorthand can be given, multiple marks can be drawn by passing an array of names or shorthands. When `none` no marks will be drawn. A style `dictionary` can be given instead of a `string` to override styling for that particular mark, just make sure to still give the mark name using the `symbol` key otherwise nothing will be drawn!.
+  ],
 )
```
Enter-tainer, line 341: Only the right bracket is pushed into next line. This might not be intended? Perhaps we should keep it if only one side has space

Enter-tainer (approved):
overall LGTM

tests/snapshots/assets__check_file@cetz-manual.typ-40.snap
QuadnucYard: I think the writer is to blame for this extra space

QuadnucYard: We should reach a consensus on this.
If the extra space is a mistake, breaking the line can help the write notice the mistake.
I feel it like a strange flavor when choosing to keep it with one-side space.

QuadnucYard:
A special case not considered.

```typ
#[// comments
foo]
```

It seems that there is no leading space. But actually, the linebreak between the line comment and the text accounts for it.

src/pretty/markup.rs
```diff
@@ -47,10 +132,116 @@ impl<'a> PrettyPrinter<'a> {
...
+        .items
+        .last()
+        .is_some_and(|last| last.node.kind() == SyntaxKind::Space)
+    {
+        items.items.pop();
+        items.has_trailing_space = true;
+    }
+
+    // Check space inside comments
+    if !items.has_leading_space
```
Enter-tainer, line 226: is it possible to fuse comment related check into previous loop? i.e. we can loop through each children and ignore comment nodes.

Enter-tainer (approved):
overall LGTM

PR #183, opened 2024-12-15
ci: better bench report

Enter-tainer:
Reduce benchmark time to 1s

Add latest benchmark status to comment


crates/typstyle-core/benches/pretty_print.rs
```diff
@@ -59,8 +59,9 @@ fn benchmark_pretty(c: &mut Criterion) {
 criterion_group! {
     name = benches;
     config = Criterion::default()
-        .measurement_time(Duration::from_secs(8))  // 只增加测量时间
-        .sample_size(50);                          // 减少采样数量
+        .warm_up_time(Duration::from_secs(1))
+        .measurement_time(Duration::from_secs(1))
+        .sample_size(30);
     targets = benchmark_attrs, benchmark_pretty
```
Enter-tainer, line 65: Seems that criterion has its own idea of sampling. So code change here is not useful😅

QuadnucYard (approved):

PR #185, opened 2024-12-15
test: organize fixtures and snapshot files

QuadnucYard:
Move `.typ` in `assets` to `fixtures`, and update meta in snaps. 

To avoid enormous diff given by GitHub PR, split the change into two PRs.

Since we did not name benchmark targets well, this PR will break benchmark results due to renaming.


Enter-tainer:
What will the second PR do? Current status LGTM

Enter-tainer (approved):

PR #188, opened 2024-12-16
fix: escape hatch not effective in some cases

### LINKED ISSUE #182: escape hatch not respected inside `if` statement
Cross-posting as requested from https://github.com/Myriad-Dreamin/tinymist/issues/996

According to the [docs](https://github.com/Enter-tainer/typstyle?tab=readme-ov-file#escape-hatch):
> If you find typstyle is not working as expected, you can use // @typstyle off or /* @typstyle off */ to disable the formatter on the next node of code.

However, neither comment style prevents tinymist from applying formatting in certain circumstances.

```typst
#if true {
  // @typstyle off
  let x = (
  "A message that will be split into multiple lines to ensure that each"
  + " line does not exceed sixty characters in length. This is a common"
  + " practice to improve readability and maintainability of the code."
  )
  // @typstyle on
}
```
Without the if block, the escape hatch works. Inside if, it fails.

_Side note: I think splitting operators to their own lines is a "favorable" default behavior compared to merging them into one line when the result would be very long_

QuadnucYard:
Fixes #182:

- If `@typstyle off` appears before the first expression in a code block, it will not take effect, since this comment is not inside the `Code` that expressions lie in.
- `@typstyle off` does not apply to the expression after hash (`#`).

### DIFF
--- a/crates/typstyle-core/src/attr.rs
+++ b/crates/typstyle-core/src/attr.rs
@@ -156,7 +156,7 @@ impl AttrStore {
                 continue;
             }
             // no format nodes with @typstyle off
-            if child_kind != SyntaxKind::Space && disable_next {
+            if disable_next && !matches!(child_kind, SyntaxKind::Space | SyntaxKind::Hash) {
                 self.set_format_disabled(child);
                 disable_next = false;
                 continue;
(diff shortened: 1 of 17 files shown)


crates/typstyle-core/src/attr.rs
```diff
@@ -158,7 +158,9 @@ impl AttrStore {
             // no format nodes with @typstyle off
             if child_kind != SyntaxKind::Space && disable_next {
```
Enter-tainer, line 159: should we move the hash check here?

Enter-tainer (approved):

crates/typstyle-core/src/attr.rs
QuadnucYard: this means `#` can be format-disabled. move hash check to the condition is also ok.

PR #189, opened 2024-12-16
feat: do not break short dot chains

QuadnucYard:
If a dot chain (without comments) starts with an ident, ends with a func call (no other calls inside), and the chain length (estimated from idents length) does not exceed 60% of max-width, the idents will not break into multiple lines.

### DIFF
--- a/crates/typstyle-core/src/pretty/code_chain.rs
+++ b/crates/typstyle-core/src/pretty/code_chain.rs
@@ -1,3 +1,4 @@
+use itertools::Itertools;
 use pretty::DocAllocator;
 use typst_syntax::{ast::*, SyntaxKind, SyntaxNode};
 
@@ -5,6 +6,7 @@ use crate::PrettyPrinter;
 
 use super::{
     chain::{iterate_deep_nodes, ChainStyle, ChainStylist},
+    util::has_comment_children,
     ArenaDoc,
 };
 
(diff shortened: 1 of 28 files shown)


Enter-tainer:
> the chain length (estimated from idents length) does not exceed 60% of max-width

Should we consider indentions here? I guess it's not covered at this moment but I think it's ok at least for now.

crates/typstyle-core/src/pretty/code_chain.rs
```diff
@@ -21,25 +27,68 @@ impl<'a> PrettyPrinter<'a> {
...
+
+        // If the chain idents are long enough, do not put them in one line.
+        let estimated_len = ident.get().len()
+            + chain
+                .iter()
+                .skip(1)
+                .map(|child| {
+                    child
+                        .cast::<FieldAccess>()
+                        .map(|field_access| field_access.field().get().len() + 1)
```
Enter-tainer, line 74: `field().len() + 1` 🤔, Are we calculating the length of `.` here ?

crates/typstyle-core/src/pretty/code_chain.rs
```diff
@@ -21,25 +27,68 @@ impl<'a> PrettyPrinter<'a> {
...
+    ) -> Option<ArenaDoc<'a>> {
+        chain.reverse();
+        let func_call = chain.last()?.cast::<FuncCall>()?;
+        let ident = chain[0].cast::<Ident>()?;
+
+        // If the chain idents are long enough, do not put them in one line.
+        let estimated_len = ident.get().len()
+            + chain
+                .iter()
+                .skip(1)
```
Enter-tainer, line 70: quick question: which item are we skipping?

QuadnucYard: the starting ident

QuadnucYard:
> Should we consider indentions here? I guess it's not covered at this moment but I think it's ok at least for now.

No. So the solution is context-unaware. As long as the chain is estimated to be able to lie in a line, it will not break.

Enter-tainer (approved):
LGTM

PR #193, opened 2024-12-18
fix: preserve comments in math

QuadnucYard:


crates/typstyle-core/src/macros.rs
```diff
@@ -0,0 +1,8 @@
+#[macro_export]
+macro_rules! alternate {
```
Enter-tainer, line 2: Could you elaborate the name here?

Enter-tainer (approved):
thx

PR #194, opened 2024-12-18
refactor: create Typstyle with config and get tab spaces from config

QuadnucYard:
Although we do not mean to expose `tab_spaces` to CLI users at the moment, it is fine to use configurable values instead of magic numbers. This can also benefit lib users.

Considering that we may add more printing methods to `Typstyle` (in the upcoming partial formatting), which all require configs, I think it is better to replace the existing `width` param with config for extensibility. This is an API break.

### DIFF
--- a/crates/typstyle-core/src/pretty/mod.rs
+++ b/crates/typstyle-core/src/pretty/mod.rs
@@ -245,9 +245,9 @@ impl<'a> PrettyPrinter<'a> {
             } else {
                 self.arena.line()
             };
-            doc = (block_sep.clone() + doc).nest(2) + block_sep;
+            doc = (block_sep.clone() + doc).nest(self.config.tab_spaces as isize) + block_sep;
         } else {
-            doc = doc.nest(2);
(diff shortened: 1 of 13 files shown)


crates/typstyle-core/src/pretty/mod.rs
```diff
@@ -368,7 +371,7 @@ impl<'a> PrettyPrinter<'a> {
                 self.arena.nil()
             },
         )
-        .nest(2)
+        .nest(self.config.tab_spaces as isize)
```
Enter-tainer, line 374: should we just define it as isize?

Enter-tainer (approved):

crates/typstyle-core/src/pretty/mod.rs
QuadnucYard: The isize is required by pretty, where dedent is allow. 
But in the config, a signed number of spaces is meaningless, IMO.

Enter-tainer: make sense

PR #195, opened 2024-12-20
feat: emit error if source is erroneous instead of silence

QuadnucYard:
- (API change) Now `Typstyle::pretty_print` returns `Result<String, FormatError>` instead of `String`.  We will have similar APIs in the upcoming partial formatting. WASM API is unchanged.
- (fix) Compilation error with WASM API introduced in #194.
- (test) Input with errors is marked erroneous in info.
- (refactor) Organize CLI format code.
- (perf) Avoid writing back if unchanged. Avoid cloning the original text if it is erroneous.

Input with error is kept accepted in check mode.
We noticed that files are still formatted even in check mode. This will be handled in the next PR.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4070,6 +4070,7 @@ dependencies = [
  "libtest-mimic",
  "pathdiff",
  "pretty_assertions",
+ "typst-syntax",
  "typstyle-consistency",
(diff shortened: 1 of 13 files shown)


Enter-tainer (approved):
LGTM

PR #198, opened 2024-12-21
fix: cli check behavior and format-all args

QuadnucYard:
- (feature) Add simple logging. Verbosity can be customized.
- (feature) Validate args before execution. `--inplace` conflicts with `--check`. `--inplace` with no input is rejected.
- (feature) Display duration in `format-all` command.
- (fix) Run with `--check` now will not modify files.
- (fix) `format-all` command now correctly has `--column` and `--check` args.
- (fix) Missing readme in crates.
- (chore) Update CLI usage in README.


QuadnucYard:
We can add optional rayon support later (if needed).

I tested it on my PC. To format ~7700 files in typst/packages, it takes 5.5s without rayon, and 1.45s with rayon.

Enter-tainer (approved):
thx

PR #202, opened 2024-12-22
feat: basic partial formatting support

QuadnucYard:
Progress:

- [x] Adjust API to allow for better integration with tinymist (subject to change before reaching a consensus).
- [x] Avoid rewriting formatting logic in CLI.
- [x] Provide basic API to format node in range. 
- [x] Add relevant tests.
- [ ] ~Provide CLI options for range formatting.~

References:

- Format API in ruff.


crates/typstyle-core/src/partial.rs
```diff
@@ -0,0 +1,126 @@
...
+        let indent = utils::count_spaces_after_last_newline(source.text(), range.start);
+        let res = doc
+            .nest(indent as isize)
+            .pretty(self.config.max_width)
+            .to_string();
+        Ok((range, res))
+    }
+}
+
+pub fn utf16_range_to_byte_range(
```
Myriad-Dreamin, line 44: We don't have to process the utf-16 encoding in typstyle. The other programs like tinymist will preprocess utf-16 things for typstyle.

crates/typstyle-core/src/pretty/config.rs
```diff
@@ -1,36 +1,36 @@
 /// Configuration Options for Typstyle Printer.
-#[derive(Debug, Clone)]
-pub struct PrinterConfig {
+#[derive(Debug, Clone, PartialEq, Eq, Hash)]
+pub struct TypstyleConfig {
```
Myriad-Dreamin, line 3: remove the useless prefix `Typstyle` in `TypstyleConfig`, becoming simply `Config`.

crates/typstyle-core/src/pretty/config.rs
```diff
@@ -1,36 +1,36 @@
 /// Configuration Options for Typstyle Printer.
-#[derive(Debug, Clone)]
-pub struct PrinterConfig {
+#[derive(Debug, Clone, PartialEq, Eq, Hash)]
```
Myriad-Dreamin, line 2: Should we derive serde? Will typstyle finally have configuration files?

crates/typstyle-core/src/partial.rs
```diff
@@ -0,0 +1,126 @@
...
+
+use crate::{pretty::Mode, utils, AttrStore, FormatError, PrettyPrinter, Typstyle, TypstyleConfig};
+
+impl Typstyle {
+    /// Format the node with minimal span that covering the given *byte* range.
+    /// If you have a utf16 range instead, use [`utf16_range_to_byte_range`] to construct the byte range.
+    pub fn format_byte_range(
+        &self,
+        source: &Source,
+        byte_range: Range<usize>,
```
Myriad-Dreamin, line 16: rename `byte_range` to `utf8_range` and then we can strip out comment "If you have a utf16 range instead, use [`utf16_range_to_byte_range`] to construct the byte range."

crates/typstyle-core/src/partial.rs
```diff
@@ -0,0 +1,126 @@
...
+    ast::{Expr, Markup},
+    LinkedNode, Source, Span, SyntaxKind,
+};
+
+use crate::{pretty::Mode, utils, AttrStore, FormatError, PrettyPrinter, Typstyle, TypstyleConfig};
+
+impl Typstyle {
+    /// Format the node with minimal span that covering the given *byte* range.
+    /// If you have a utf16 range instead, use [`utf16_range_to_byte_range`] to construct the byte range.
+    pub fn format_byte_range(
```
Myriad-Dreamin, line 13: `format_byte_range` -> `format_source_range`, since we have `format_source`.

QuadnucYard:
Should we provide line-column to range API in typstyle-core?
Or just use it in CLI?

crates/typstyle-core/src/pretty/config.rs
Enter-tainer: We may add a feature flag to control this. Like `serde` will enable it. For config files, I'm not willing to add it at this moment.

Enter-tainer:
> Should we provide line-column to range API in typstyle-core? Or just use it in CLI?

We may add minimal API that works with tinymist in the first prototype.

crates/typstyle-core/src/lib.rs
```diff
@@ -27,63 +27,61 @@ impl std::fmt::Display for FormatError {
...
-    ///     max_width: 80,
-    ///     ..Default::default()
-    /// }).pretty_print();
-    /// ```
-    pub fn new_with_content(content: impl Into<String>, config: PrinterConfig) -> Self {
+    pub fn new() -> Self {
+        Default::default()
+    }
+
+    pub fn config(mut self, config: Config) -> Self {
```
Enter-tainer, line 40: maybe `with_config`

crates/typstyle-core/src/lib.rs
```diff
@@ -27,63 +27,61 @@ impl std::fmt::Display for FormatError {
...
+        self.config = config;
+        self
+    }
+
+    pub fn config_width(mut self, max_width: usize) -> Self {
```
Enter-tainer, line 45: maybe user should take a `&mut` of config and edit it directly? I'm not sure but this interface doesn't look good to me. Or we may forbid user from editing config after typstyle is constructed.

Enter-tainer: cc @Myriad-Dreamin

crates/typstyle-core/src/lib.rs
```diff
@@ -27,63 +27,61 @@ impl std::fmt::Display for FormatError {
...
-            config,
-        }
+    pub fn format_source(self, source: &Source) -> Result<String, FormatError> {
+        self.format_source_inspect(source, |_| {})
     }
 
-    /// Pretty print the content to a string.
-    pub fn pretty_print(&self) -> Result<String, FormatError> {
-        let root = self.source.root();
+    pub fn format_source_inspect(
```
Enter-tainer, line 59: cool

crates/typstyle-core/src/partial.rs
```diff
@@ -0,0 +1,75 @@
...
+        let doc = if let Some(markup) = node.cast() {
+            printer.convert_markup(markup)
+        } else if let Some(expr) = node.cast() {
+            printer.convert_expr(expr)
+        } else if let Some(pattern) = node.cast() {
+            printer.convert_pattern(pattern)
+        } else {
+            return Err(FormatError::SyntaxError);
+        };
+        let indent = utils::count_spaces_after_last_newline(source.text(), range.start);
```
Enter-tainer, line 37: interesting..

crates/typstyle-core/src/lib.rs
Myriad-Dreamin: Reconfiguration can happen when hot reloading lsp configurations but I'm not sure whether it is good to expose a `&mut config` directly because you may derive and store some value from config internally and it will cause inconsistency to change config by the exposed mut.

Enter-tainer: Currently we don't have extra state in config. If that's needed, we might need to change the struct hierarchy here. Maybe we can use a simple design at this moment and revisit it when we need more functions.

BTW I just realized that currently we will create a new typstyle instance every time when we format a file. So technically it is "reconfigured" every single time.

Myriad-Dreamin (approved):

PR #205, opened 2024-12-24
fix: do not always add end-of-file newline

ramdomPTM:
Introduced in c6cae33, `typstyle` would always add a `\n` at the end of the file,  leading to the unwanted effect of an ever-extending `\n` sequence.

The expected results are:
1. if there's no `\n` at the end of the file, add one;
2. if there are more than one `\n`s, shrink to one;
3. probably remove `\n`s at the start of the file.

This PR fixes that.

### DIFF
--- a/crates/typstyle-core/src/lib.rs
+++ b/crates/typstyle-core/src/lib.rs
@@ -82,8 +82,10 @@ pub fn strip_trailing_whitespace(s: &str) -> String {
         .lines()
         .map(|line| line.trim_end())
         .collect::<Vec<_>>()
-        .join("\n");
-    res + "\n"
+        .join("\n")
+        .trim()
+        .to_owned();
+    res
 }
 
 #[cfg(all(target_arch = "wasm32", feature = "wasm"))]


Enter-tainer:
> leading to the unwanted effect of an ever-extending \n sequence.

i wonder why our tests dont catch this

Enter-tainer (approved):

Enter-tainer:
> leading to the unwanted effect of an ever-extending \n sequence.

how to reproduce this in current version?

ramdomPTM:
My current setting: Helix editor built from master and typstyle v0.12.13.

Settings for Helix:
```toml
[[language]]
name = "typst"
formatter = { command = "typstyle" }
```

Using `:fmt` in Helix to format a typst file would result in the described behavior.

https://asciinema.org/a/nuLywqrpaMzp3gPiyaktvm06x

Enter-tainer:
is it specific to your editor? does typstyle-cli have the same issue?

ramdomPTM:
It's not specific to the editor, for cli generated the same result, adding newlines no matter what.

typstyle version:
```
typstyle 
Version:             0.12.13
Build Timestamp:     2024-12-22T15:36:39.504661058Z
Build Git Describe:  VERGEN_IDEMPOTENT_OUTPUT
Commit SHA:          VERGEN_IDEMPOTENT_OUTPUT
Commit Date:         None
Commit Branch:       None
Cargo Target Triple: x86_64-unknown-linux-gnu
```

https://asciinema.org/a/aAp1QUpE9tqAFRaKiCAUGWzzJ

Myriad-Dreamin:
The bug actually happens here.

https://github.com/Enter-tainer/typstyle/blob/d265af8b7b4eccec462c7db7f9459f6079ff3ea2/crates/typstyle/src/fmt.rs#L184

QuadnucYard:
Sorry we didn't consider piping usage before.
A simple fix is just to change that `println` to `print`.
In order to pass the tests, you need to update insta snapshots in cli tests.

Your fix is actually a style change and solves this problem by coincidence. We have previously discussed whether to keep trailing linebreaks. There is always someone who prefers free blank lines at the end of the document. We don't know whether stripping is a preferable style done by typstyle. cc @Enter-tainer 

btw, if you want to keep your change, please rebase on the latest and update snapshots in tests.

ramdomPTM:
Thanks for your information! I understand that someone may prefer trailing end-of-lines. Now that the real problem has been spotted, please feel free to close this PR.

QuadnucYard:
If you are willing to contribute, feel free to open another PR to fix the printing! (remember to update snapshots)

PR #207, opened 2024-12-25
Revert "ci: add build attest"

Enter-tainer:
Reverts Enter-tainer/typstyle#204


Enter-tainer:
We should only enable it for releases. But currently I don't have time to maintain it.

PR #208, opened 2024-12-26
fix: do not add linebreak at the start of the document when there is no parbreak

QuadnucYard:


crates/typstyle-core/src/pretty/markup.rs
```diff
@@ -12,6 +12,9 @@ use super::{
 
 #[derive(Default)]
 struct MarkupStyle {
+    /// Whether this markup is in content block.
+    pub in_block: bool,
+    /// Whether the spaces can be safely stripped.
     pub strip_space: bool,
```
Enter-tainer, line 18: `can_strip_space` maybe

crates/typstyle-core/src/pretty/markup.rs
```diff
@@ -12,6 +12,9 @@ use super::{
 
 #[derive(Default)]
 struct MarkupStyle {
+    /// Whether this markup is in content block.
+    pub in_block: bool,
```
Enter-tainer, line 16: consider `is_top_level`, as all other markup should be in block

Enter-tainer (approved):

Issue #122, opened 2024-07-23
Define number of indent spaces.


PhosCity:
Is there a way for me to define the number of indent spaces in typstyle? I wanted the number of indent space to be 4 but typstyle defaults at 2 spaces.

Enter-tainer:
i believe this is a deplicate of #116

PhosCity:
It is! In my defense, I did search for issues about indents and didn't find any but did not search for issues about tabs. I'll close this issue and follow that one.

PR #220, opened 2025-02-10
Add `--tab-width`, which sets number of spaces for levels of indentation in the output

### LINKED ISSUE #116: Custom tab size
I know that the canonical tabbing for Typst is "two spaces", but for a lot of users that's unusual. Thus, for all languages that recommend two-space, I (and a lot of programmers I know) use a four-space tabbing during development, and then turn it into two-space right before publishing, often automatically. 

Because of that, I propose having a setting for tab size. I imagine it could be easy. As a last resort, it could be crutched as 
`doNormalFormatting(text).replace("\n(  )+") { "\n" + " ".repeat(it.length / 2 * size) }` or whatever its analogue in Rust is...

AgentElement:
This PR closes #116 and #122.

Although `typstyle` is opinionated, tab width is a common setting to expose to the user in other opinionated formatters.

### DIFF
--- a/crates/typstyle-core/src/config.rs
+++ b/crates/typstyle-core/src/config.rs
@@ -30,6 +30,11 @@ impl Config {
         self
     }
 
+    pub fn with_tab_spaces(mut self, tab_spaces: usize) -> Self {
+        self.tab_spaces = tab_spaces;
+        self
+    }
+
     pub fn chain_width(&self) -> usize {
         const CHAIN_WIDTH_RATIO: f32 = 0.6;
         (self.max_width as f32 * CHAIN_WIDTH_RATIO) as usize
--- a/crates/typstyle/src/cli.rs
+++ b/crates/typstyle/src/cli.rs
@@ -74,6 +74,16 @@ pub struct StyleArgs {
         help_heading = "Format Configuration"
     )]
     pub column: usize,
+
+    /// Spaces per level of indentation in the output
+    #[arg(
+        short,
+        long,
+        default_value_t = 2,
+        global = true,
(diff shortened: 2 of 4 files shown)


Enter-tainer (approved):
thanks for adding this!

Enter-tainer:
BTW, could you please add a cli e2e test for tab width flag? I remember we have a snapshot based cli test framework

AgentElement:
Done!

QuadnucYard:
Could you please rebase on master to make the test pass?

AgentElement:
Just rebased.

QuadnucYard:
Sorry, please update the snapshot. I just changed the output format a bit.

AgentElement:
Rebased again.

QuadnucYard:
thanks!

AgentElement:
Apologies for necrobumping, but would you be willing to cut a new release with this change? I'd like to add a config parameter to tinymist to set tab width. That would require typstyle to have a version bump (I don't want to introduce a git repo as a dependency for tinymist).

Enter-tainer:
i will release it in the weekend.

AgentElement:
> Does this PR add support for using tabs instead of spaces?

It does not.

QuadnucYard:
> Does this PR add support for using tabs instead of spaces?

The internally used `pretty.rs` unfortunately does not support tabs.

But we can simply replace leading spaces by tabs in the final result if needed:

https://github.com/Enter-tainer/typstyle/blob/43ff4e3cbf199964017942014e9a44404cf24a36/crates/typstyle-core/src/utils.rs#L9-L12

PR #221, opened 2025-02-11
fix: test failure caused by output order and files plural

QuadnucYard:
When formatting files in a directory, the order is actually undefined.
Currently, there is no way to sort the output order in the cli test, except sorting the files beforehand.

Simply reduce the files to format in the failed test to make it pass.


Enter-tainer (approved):
thx

PR #225, opened 2025-02-15
dev: rework markup boundary to fix list/enum/term item indent with comments

### LINKED ISSUE #216: Bug: formatting nested enum/list
Typstyle changes:
```typst
+
  + xyz
```
which creates a nested enum into:
```typst
+ + xyz
```
which is not parsed as a nested enum by typst.

It also transforms:
```typst
+ //foo
  + xyz
```
into:
```typst
+ //foo
+ xyz
```
which changes the item level.

Nested unordered lists have the same issue.

QuadnucYard:
Now it should correctly handle nested lists and comments mixed in.
Fixes #216. Leading linebreaks are no longer stripped.

This may still need to change in 0.13 due to syntax refactoring.


Enter-tainer (approved):
LGTM

PR #233, opened 2025-02-26
feat: use braces instead of parentheses for multiline closure bodies

QuadnucYard:
Use braces (`{}`) instead of parens (`()`) when the closure body is not a chainable binary expression. e.g.

```typst
#fun(() => {
  aaa
})
instead of
#fun(() => (
  aaa
))
```

Braces are more intuitive and align with the behavior of other languages.
I exclude binary expressions to avoid double nesting.

PERFORMANCE CONSIDERATION: The code block has extra evaluation burden.


Enter-tainer (approved):
cool

PR #241, opened 2025-03-19
fix: disable format for import with comments to avoid mis-format

### LINKED ISSUE #239: Import item with comment is misformatted
```
#import "@preview/jogs:0.2.4": (
  // test
  eval-js)
```

Enter-tainer:
fix #239

### DIFF
--- a/crates/typstyle-core/src/attr.rs
+++ b/crates/typstyle-core/src/attr.rs
@@ -1,7 +1,7 @@
 use rustc_hash::FxHashMap;
 
 use typst_syntax::{
-    ast::{Args, AstNode, Math, Raw},
+    ast::{Args, AstNode, Math, ModuleImport, Raw},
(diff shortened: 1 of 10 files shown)


crates/typstyle-core/src/attr.rs
```diff
@@ -171,6 +171,10 @@ impl AttrStore {
         if commented {
             self.set_commented(node);
         }
+        if commented && node.cast::<ModuleImport>().is_some() {
+            // no format module import if it has comments inside
+            self.set_format_disabled(node);
+        }
```
Enter-tainer, line 177: This is unfortunate because i didn't come up with a good method to handle comments here

tests/fixtures/unit/code/import-comment.typ
Enter-tainer, line 1: maybe i should move this to `in-import.typ`

Issue #212, opened 2024-12-29
Feature Proposal: Grouped Import Sorting in Typstyle (Similar to Python’s isort)


Enter-tainer:
**generated by genai from unstructured chat logs**

#### Issue Description
Currently, Typstyle does not have any functionality for sorting imports. This can lead to inconsistent and disorganized import statements in code, reducing readability and maintainability. This proposal suggests adding **import sorting** to Typstyle, similar to tools like Python’s `isort`, to automatically organize imports in a consistent and logical manner.

#### Proposed Feature
Implement an **import sorting** feature that:
1. **Sorts Imports Alphabetically**: Orders `import` statements alphabetically.
2. **Groups Imports (Optional)**: Allows imports to be grouped logically (e.g., standard library, third-party, local modules) with empty lines separating groups.

#### Implementation Considerations
1. **Basic Sorting**:
   - Sort all imports alphabetically by default.
   - Handle both `import` and `from ... import` statements.
2. **Grouping (Optional)**:
   - Detect and group imports by type (e.g., standard library, third-party, local modules).
   - Separate groups with empty lines.

cc @QuadnucYard

PR #246, opened 2025-03-26
feat: reorder import items if no comment or duplication

QuadnucYard:
Add an option to reorder import items (default: false). Part of #212.

We sort the items of an import statement alphabetically, if it has no comment or duplicated name. In other cases, the order should not matter.

We designed a directive to set local configuration in different unit tests: `/// typstyle: some-options`.

### DIFF
--- a/tests/fixtures/unit/code/snap/import-reorder.typ-120.snap
+++ b/tests/fixtures/unit/code/snap/import-reorder.typ-120.snap
@@ -1,0 +1,58 @@
+---
+source: tests/src/unit.rs
+input_file: tests/fixtures/unit/code/import-reorder.typ
+---
+/// typstyle: reorder-import-items
+
+// Single import with mixed styles in random order
+#import "module.typ": (
+  a.b.c,
+  alpha,
+  beta,
+  func as renamed,
+  gamma,
+  h.i.j,
+  j.k.l as util,
+  m.n,
+  p.q as custom,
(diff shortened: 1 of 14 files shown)


tests/fixtures/unit/code/snap/import-reorder.typ-80.snap
```diff
@@ -0,0 +1,64 @@
+---
+source: tests/src/unit.rs
+input_file: tests/fixtures/unit/code/import-reorder.typ
+---
+/// typstyle: reorder-import-items
```
Enter-tainer, line 5: should we use `@typstyle` to keep consistency between `@typstyle off`?

tests/fixtures/unit/code/snap/import-reorder.typ-120.snap
```diff
@@ -0,0 +1,58 @@
...
+  // This comment should prevent reordering
+)
+#import "commented.typ": ddd, /* Another comment type */ ccc
+
+// Case 2: Import with duplicate item names - should not reorder
+// Same name imported twice
+#import "duplicates.typ": helper, widget, helper, formatter
+
+// Same name imported directly and via alias
+#import "alias-duplicates.typ": config, settings as config, template
```
Enter-tainer, line 55: Is this valid??

Enter-tainer (approved):
cool

PR #247, opened 2025-03-27
feat: format args and hashed expressions in math

### LINKED ISSUE #22: Correctly handle function calls in math mode
We refuse to format them currently. It's very difficult

### LINKED ISSUE #132: bad indent for math equation
```

#import "@preview/physica:0.9.0": *
#import "@preview/ctheorems:1.1.2": thmrules, thmbox


#let problem = thmbox(
  "problem",
  "Solution to Problem",
)
#show: thmrules

#problem("1")[
  $
    pdv(,t)braket(phi.alt (t), psi(t))=hbar / i
    mel(
    phi.alt(t),
    (-H^dagger (hat(x),hat(p), t)+H(hat(x),hat(p),t))
    ,psi(t)
  ),
  $
]
```

QuadnucYard:
Closes #22. Fixes #132 and #141.

There are some problems:
1. It is space-sensitive, and there may be some content magic on the user side. So we should not deliberately change spaces in math.
2. We cannot distinguish between `sin` and `cases`. They have the same AST, but only the latter is really a function. Therefore, we must keep commas in args and not add extra ones.

These problems mean we have to keep the structure.

Line-breaking in math is also suppressed.


Enter-tainer:
interestingly this pr greatly improves the performance🤔

Enter-tainer (approved):
snapshot LGTM

PR #252, opened 2025-04-06
fix: correct handling comments directly in equations and spaces in math

QuadnucYard:
Fixed some bugs introduced in #247 and #248.

- Correctly handles in-equation comments before and after `Math`.
- Fixed infinite spaces added before `//` in math.
- Fixed extra indent of array args in math.
- Eats spaces in `MathAttach` and `MathRoot`.


crates/typstyle-core/src/pretty/code_list.rs
```diff
@@ -103,6 +103,8 @@ impl<'a> PrettyPrinter<'a> {
                 add_trailing_sep_single: is_explicit,
                 add_trailing_sep_always: ends_with_comma,
                 delim: if is_explicit { ("(", ")") } else { ("", "") },
+                tight_delim: !is_explicit,
```
Enter-tainer, line 106: quick question: why implicit lists should not be indented and should have tight delim?

QuadnucYard: > quick question: why implicit lists should not be indented and should have tight delim?

to prevent implicit arrays in math args from being formatted as
```typ
$
mat(
  1, 2;
    3333,
    4444,
  ; 5, 6
)
$
```

Enter-tainer (approved):

PR #255, opened 2025-04-10
feat: unify equation layout and don't indent non-block equations

QuadnucYard:
- Don't indent non-block equations. This is expected to work well in most reasonable cases.
- Treat nodes with multiline comments as multiline.
- Fixed some issues with indent and grouping in list layout.
- Improved comment attachment.
- Fixed missing separators in list layout mixed with comments in some cases.


crates/typstyle-core/src/pretty/list.rs
```diff
@@ -5,35 +5,38 @@ use crate::ext::StrExt;
...
-    Linebreak(usize),
-}
-
 pub struct ListStylist<'a> {
     printer: &'a PrettyPrinter<'a>,
     can_attach: bool,
     free_comments: Vec<ArenaDoc<'a>>,
     peek_hash: bool,
     items: Vec<Item<'a>>,
-    item_count: usize,
```
Enter-tainer, line 28: do we have fake items by the way?

Enter-tainer (approved):

crates/typstyle-core/src/pretty/list.rs
QuadnucYard: > do we have fake items by the way?

Naming is a tricky matter... The `item` here refers to anything that can make up a line, including comments and hard linebreaks. Trivia is not counted as `real_item`. I don't have a good naming idea though.

PR #257, opened 2025-04-15
refactor: collect markup lines to allow for line-based actions

QuadnucYard:
Performance considerations:
A large markup expression can consist of many lines, but each line may contain only a few nodes. We may adopt a more efficient structure.


Enter-tainer (approved):
LGTM

PR #260, opened 2025-04-17
feat: wrap text in markup

### LINKED ISSUE #104: reflow comment and maybe content
https://github.com/Enter-tainer/typstyle/issues/103#issuecomment-2212481529

QuadnucYard:
Optional feature enabled by `--wrap-text`.

Closes #104. Handling contents in comments are yet another challenging topic, so we will not consider reflowing comments at the time.

~~Note: since we have some context-sensitive decisions, such as deriving fold style from the location in a paragraph, it may not converge until a few passes with wrapping enabled.~~

To ensure convergence on the first pass, and improve output quality, in addition not to suppress line breaks in markup, we will:
- Turn spaces around block equations into hardline.
- *Not* turn linebreaks around the block elements (e.g., headings, lists), codes following behind hashes or standalone non-text elements, into spaces.


tests/fixtures/packages/snap/cetz-manual.typ-0.snap
```diff
@@ -660,7 +660,9 @@ line((0,0), (30deg, 1), (60deg, 1),
 ```
 
 == Barycentric
-In the barycentric coordinate system a point is expressed as the linear combination of multiple vectors. The idea is that you specify vectors $v_1$, $v_2$ ..., $v_n$ and numbers $alpha_1$, $alpha_2$, ..., $alpha_n$. Then the barycentric coordinate specified by these vectors and numbers is $ (alpha_1 v_1 + alpha_2 v_1 + dots.c + alpha_n v_n) / (alpha_1 + alpha_2 + dots.c + alpha_n) $
+In the barycentric coordinate system a point is expressed as the linear combination of multiple vectors. The idea is that you specify vectors $v_1$, $v_2$ ..., $v_n$ and numbers $alpha_1$, $alpha_2$, ..., $alpha_n$. Then the barycentric coordinate specified by these vectors and numbers is $
```
Enter-tainer, line 663: If i understand it correctly, 0 width tests should have exactly 1 word per line?

Enter-tainer: or it's not enabled in tests yet?

tests/src/unit.rs
```diff
@@ -124,27 +126,47 @@ fn check_snapshot(path: &Path, width: usize) -> Result<(), Failed> {
...
     cfg.max_width = width;
-    let first_pass = Typstyle::new(cfg.clone()).format_source(&source)?;
-    let second_pass = Typstyle::new(cfg).format_content(&first_pass)?;
-    pretty_assertions::assert_str_eq!(
-        first_pass,
-        second_pass,
-        "first pass and second pass are not the same!"
-    );
+    let mut first_pass = Typstyle::new(cfg.clone()).format_source(&source)?;
+    for i in 0..=opt.relax_convergence {
```
Enter-tainer, line 137: interesting

Enter-tainer:
About relaxed convergence, should we format the files repeatedly until it converges?

tests/fixtures/packages/snap/cetz-manual.typ-0.snap
QuadnucYard: > or it's not enabled in tests yet?

Only enabled by the directive.
This feature should be tested on more large documents.
I think we can add a file to describe additional testing configurations for certain source files, in a non-intrusive manner, so that we can reuse sources.

Enter-tainer:
> ignore the structural information to make it converge, but the output may look ugly.

Could you please elaborate?

QuadnucYard:
> > ignore the structural information to make it converge, but the output may look ugly.
> 
> Could you please elaborate?

Normally, we tend to fold when the code or equation is mixed in texts. But when text-wrapping is enabled, its context may change and no longer lie in texts. This means it possibly reflows again in another format pass.

Now I have two countermeasures:
- When text-wrapping is enabled, ignore the fact "whether the piece of code is mixed in texts".
- (new) Do not turn the linebreaks around a piece of code into spaces in reflow, when the code takes up a whole line. This may still need multiple passes before converging, but it will definitely converge.

crates/typstyle/src/cli.rs
```diff
@@ -66,65 +66,40 @@ pub enum Command {
...
-        long,
-        default_value_t = false,
-        global = true,
-        help_heading = "Format Configuration"
-    )]
+    #[arg(long, default_value_t = false, global = true)]
     pub reorder_import_items: bool,
+
+    /// Whether to wrap texts in the markup.
+    #[arg(long, default_value_t = false, global = true)]
```
Enter-tainer, line 81: how about make if false by default now? I'm okay with `reorder_import_items = true`. But I'm a little bit concerned about text wrap

Enter-tainer (approved):

PR #261, opened 2025-04-19
feat: compact layout for args of func_call

### LINKED ISSUE #95: Better handling for last arg in function call

QuadnucYard:
Resolves #95.
Compact layout is preferred when the args does not contain comments or linebreaks, and the last arg is not binary expression.
Now the flavor of args is ignored. Otherwise, the compact layout will not apply to existing multiline args.
~~Added a width limit for the initial args except the last, which is a hardcoded 80% width.~~ This does not work well with inline chains

Some problems:
- We use `union` and a brute-force `flatten`, which are not performant.
- The trailing args in brackets are not counted in the compact layout, so they could exceed the width limit in the new layout.
  - It may look better to break inside the brackets than args. But unfortunately, we cannot unconditionally insert breaks in markup block.
- If it has two args of the same structure, they may not be formatted the same way.


QuadnucYard:
problems with compact layout:

- It does not respect the flavor of args, which is more opinionated. Can this lead to bad output?
- It does not work well with some DSLs that contain linebreaks in args. But in this case, we can simply insert a line comment to disable compact layout.

tests/fixtures/packages/snap/cetz-tree.typ-80.snap
```diff
@@ -99,10 +96,10 @@ snapshot_kind: text
         draw.content((), [?], name: "content")
       }
       if draw-node == "rect" {
-        draw.rect(
-          (rel: (-.1, .1), to: "content.top-left"),
-          (rel: (.1, -.1), to: "content.bottom-right"),
-        )
+        draw.rect((rel: (-.1, .1), to: "content.top-left"), (
```
QuadnucYard, line 99: breaking dsl

tests/fixtures/packages/snap/cetz-tree.typ-40.snap
```diff
@@ -287,10 +284,8 @@ snapshot_kind: text
       node-position(node),
     )
     for child in node.children {
-      d += anchors(
-        child,
-        parent-path + str(node.n),
-      )
+      d += anchors(child, parent-path
+        + str(node.n))
```
QuadnucYard, line 288: should not be compact when the last arg is binop

tests/fixtures/packages/snap/fletcher-diagram.typ-80.snap
```diff
@@ -499,93 +499,98 @@ input_file: tests/fixtures/packages/fletcher-diagram.typ
...
-
-
-    // determine diagram's elastic grid layout
-    let grid = compute-grid(
-      rects-affecting-grid,
-      vertices-affecting-grid,
-      options,
-    )
+  box(
+    context {
```
QuadnucYard, line 503: it seems that the long lines inside prevent it from being compact

tests/fixtures/packages/snap/fletcher-draw.typ-80.snap
```diff
@@ -23,16 +22,18 @@ snapshot_kind: text
...
+          cetz.draw.translate(node.pos.xyz)
+          for (i, extrude) in node.extrude.enumerate() {
+            cetz.draw.set-style(
+              fill: if i == 0 { node.fill },
+              stroke: node.stroke,
+            )
+            (node.shape)(node, extrude)
+          }
+        },
+      )
```
QuadnucYard, line 36: why not compact?

tests/fixtures/packages/snap/fletcher-draw.typ-80.snap
```diff
@@ -67,18 +68,13 @@ snapshot_kind: text
 
   if debug >= 2 and node.radius != 0pt {
     // node bounding rectangle
-    cetz.draw.rect(
-      ..rect-at(node.pos.xyz, node.size),
-      stroke: DEBUG_COLOR + .1pt,
-    )
+    cetz.draw.rect(..rect-at(node.pos.xyz, node.size), stroke: DEBUG_COLOR
+      + .1pt)
```
QuadnucYard, line 72: should prevent chaining here

tests/fixtures/packages/touying/snap/utils.typ-80.snap
```diff
@@ -313,9 +310,10 @@ snapshot_kind: text
   ) {
     self.methods.convert-label-to-short-heading
   } else {
-    (self: none, lbl) => titlecase(
-      lbl.replace(regex("^[^:]*:"), "").replace("_", " ").replace("-", " "),
-    )
+    (self: none, lbl) => titlecase(lbl
+      .replace(regex("^[^:]*:"), "")
+      .replace("_", " ")
+      .replace("-", " "))
```
QuadnucYard, line 316: chain looks bad

tests/fixtures/packages/touying/snap/utils.typ-80.snap
```diff
@@ -680,57 +676,60 @@ snapshot_kind: text
...
 
-    layout(container-size => {
-      // Helper function to more easily grab absolute units
-      let get-pts(body, w-or-h) = {
-        let dim = if w-or-h == "w" {
-          container-size.width
-        } else {
-          container-size.height
+    layout(
+      container-size => {
```
QuadnucYard, line 680: not compact

QuadnucYard:
there are some weird behaviors with pretty, causing compact layout unexpectedly fail to fit
one example is
```typ
#{
  {
    {
      cetz.group(
        {
          {
            cetz.draw.set-style(
              fill: if i == 0 { node.fill },
              stroke: if i == 0 { node.fill },
            )
          }
        },
      )
    }
  }
}
```
modifying almost any piece of code can make the compact layout of `cetz.group(...)` success

QuadnucYard:
Maybe we can hold on to *sticky* braces to bypass such problems.
After all, compact layout will not make things worse. It can only degrade to the layout in the current  version.

tests/fixtures/packages/touying/snap/utils.typ-80.snap
QuadnucYard: > chain looks bad

rustfmt is the same

Enter-tainer (approved):
snapshots LGTM

crates/typstyle-core/src/config.rs
```diff
@@ -46,4 +46,9 @@ impl Config {
         const CHAIN_WIDTH_RATIO: f32 = 0.6;
         (self.max_width as f32 * CHAIN_WIDTH_RATIO) as usize
     }
+
+    pub fn args_width(&self) -> usize {
```
Enter-tainer, line 50: does this still make sense? looks like it's no longer used

QuadnucYard: it may be used later, if we could truly support variable widths.

Enter-tainer: maybe we can comment it out in this pr

PR #263, opened 2025-04-22
feat(test): reimplement the e2e&consistency test framework to support combinations and full error reporting

QuadnucYard:
Changes to e2e tests and consistency tests:
- Feature: e2e tests now require successful compilation to ensure effectiveness.
- Feature: Configure e2e repos in a TOML file instead of code. Added more popular repos.
- Feature: Allow running e2e tests for a single repo with different formatter configurations and entry points. Make `reorder-import-items` enabled by default.
- Feature: Record errors in e2e tests and print them at failure. No longer bail on the first error.
- Change: Do not compare PDFs. This can be added if it is really needed.
- Change: Do not clone repos again when they already exist. This does not affect CI, but can avoid unnecessary cloning in local development.
- Build: Use O1 in test profile. This will not significantly affect testing performance.


Enter-tainer (approved):
LGTM!

PR #266, opened 2025-04-28
feat: align math by align points

QuadnucYard:
How it works:
We collect "cells" in the aligned math, and printed them into strings. We measure their unicode widths to align them by align points (`&`, `\`).

Works in these cases:
- Has any align point
- ~~No line comment~~
- ~~No multiline cell~~
- ~~Not in args or delimited~~

Disabled when:
- Has multiline `Str` or `Raw` descendants.
- Not following a linebreak when in `MathDelimited` or `Args`. Why: this can cause misalignment.

TODO list:
- [x] ~~eat `'\n'` in spaces (?)~~
- [x] support nest align points in math-delimited
- [x] support multiline cells


QuadnucYard:
Question: in the following case,
```typ
$
  1 & 2
  3 & 4 \
  5 & 6 & 7
$
```
should we keep the linebreak between `2` and `3`?

Enter-tainer:
> Question: in the following case,
> ```typ
> $
>   1 & 2
>   3 & 4 \
>   5 & 6 & 7
> $
> ```
> should we keep the linebreak between `2` and `3`?

it's similar to table cell alignment. iirc currently we will not take the linebreak into consideration

Enter-tainer:
Few quick questions, in current implementation:

1. what would happen if a line longer than column limit after padding?
2. does it take content alignment into account? I mean the alignment points are [alternating between left and right](https://sitandr.github.io/typst-examples-book/book/basics/math/alignment.html#alignment-points).

QuadnucYard:
> 1. what would happen if a line longer than column limit after padding?

Just exceeding. Wrap cells to ensure limited width is not planned, as it is hard to adjust and may be worse.

> 2. does it take content alignment into account? I mean the alignment points are [alternating between left and right](https://sitandr.github.io/typst-examples-book/book/basics/math/alignment.html#alignment-points).

Cell alignment is also alternating.

Enter-tainer:
2 is good-to-have but I'm worrying about 1. Simply exceeding will be a mess for very long equations, especially for soft wrap users. Can we spread the equation fragments into multilines like what we do in table/grid? like

```
$
  1111111111 &
  2               &
  3333 \
  a & b & c
$
```

Or we simply disable alignment for lines that exceeds the limit

tests/fixtures/articles/snap/undergraduate-math.typ-120.snap
```diff
@@ -493,8 +493,8 @@ Definition by cases can be easily obtained with the `cases` function.
     [
       $
         f_n = cases(
-          a &"if" n = 0,
-          r dot f_(n - 1) &"else"
+          a & "if" n = 0,
+          r dot f_(n - 1) & "else"
         )
       $
```
Enter-tainer, line 499: any idea why this equation is not padded with spaces?

tests/fixtures/unit/math/aligned/commented.typ
```diff
@@ -0,0 +1,106 @@
...
+$ cases(
+  x &= y,   // First case
+  // Middle comment
+  z &= w    /* Second case */
+) &= sum x  // Right hand side
+          \
+  // Line comment between rows
+  &= product y   /* Block
+                   comment */ \
+  &= z      // Final result $
```
Enter-tainer, line 56: does this test case make sense? looks like `$` is commented out

tests/fixtures/unit/math/aligned/snap/general.typ-120.snap
```diff
@@ -0,0 +1,104 @@
...
+$
+
+// Nested alignments and complex expressions
+$
+             cases(
+         1 & 2 & 3,
+          4 & 5 & 6
+                  ) & display(sum x)  & integral_0^1 f(x) \
+  vec(1&2&3, 4&5&6) & cases(a&b, c&d) &  display(product) \
+$
```
Enter-tainer, line 25: the `cases` part doesn't looks good

tests/fixtures/unit/math/aligned/snap/nested.typ-120.snap
```diff
@@ -0,0 +1,131 @@
...
+    "Case 2: " & = integral_0^1 f(t) \
+               & = g(x)              ,
+    "Case 3:  " & = cases(
+                    a & = b \
+                      & = c ,
+                    d & = e \
+                      & = f
+                  )
+  )
+$
```
Enter-tainer, line 77: `&`s after `"Case N"` are not aligned. Is this expected?

Enter-tainer (approved):
Overall LGTM

tests/fixtures/articles/snap/undergraduate-math.typ-120.snap
QuadnucYard: oh I didn't think of alignment across args

tests/fixtures/unit/math/aligned/commented.typ
QuadnucYard: sorry, it's a mistake by copilot

QuadnucYard:
It is tricky to handle alignment across args. Let's leave it to the next PR.

Enter-tainer (approved):
LGTM

PR #271, opened 2025-05-03
feat: add method to set wrap text option in Config

Enter-tainer:
TSIA


Enter-tainer:
http://github.com/Myriad-Dreamin/tinymist/pull/1737

PR #279, opened 2025-05-09
fix: do not eat space between hashed expr and math underscore

### LINKED ISSUE #272: Incorrect formatting output
The formatter removes spaces that are essential to the correct parsing of formulas. Weirdly enough, this only happens sometimes, if the lines on which action is needed for other problems aren't formatted.

See example of a file which will trigger unwanted behavior below:

```typst
#let color_math(col, expr) = text(fill: col, weight: "bold", $bold(expr)$)
#let orange_sum = color_math(orange, $sum$)
#let blue_sum = color_math(blue, $sum$)
$
  &= sum_display(k\,l <= 2n \ k + l <= 2n) a_k b_l - underbrace(sum_(k=0)^n sum_(l=0)^n a_k b_l,"rouge") \
  &= #orange_sum _display(k\,l <= 2n \ k > n \ k + l <= 2n) a_k b_l + #blue_sum _display(k\,l <= 2n \ l > n \ k + l <= 2n) a_k b_l
$
Or
$
  & abs(#orange_sum _display(k\,l <= 2n \ k > n \ k + l <= 2n) a_k b_l) <= #orange_sum _display(k\,l <= 2n \ k > n \ k + l <= 2n) abs(a_k) abs(b_l) <= underbrace(#orange_sum _display(k\,l <= 2n \ k > n) abs(a_k) abs(b_l),"on somme des termes positifs") \
$
```

Running the formatter removes the space between `orange_sum` and `_display`, which is essential to allow the compiler to separate both objects:
Image

(Weirdly enough, if I format, then put the spaces back myself, then reformat, _it doesn't remove the spaces_)

QuadnucYard:
Fixes #272.

In tests, added a directive to bypass unnecessary testcases. Old fixtures not updated yet.

### DIFF
--- a/crates/typstyle-core/src/pretty/code_flow.rs
+++ b/crates/typstyle-core/src/pretty/code_flow.rs
@@ -11,7 +11,7 @@ use crate::ext::{BoolExt, StrExt};
 impl<'a> PrettyPrinter<'a> {
     pub(super) fn convert_named(&'a self, ctx: Context, named: Named<'a>) -> ArenaDoc<'a> {
         let mut seen_name = false;
-        self.convert_flow_like(ctx, named.to_untyped(), |ctx, child| {
+        self.convert_flow_like(ctx, named.to_untyped(), |ctx, child, _| {
             if child.kind() == SyntaxKind::Colon {
                 FlowItem::tight_spaced(self.arena.text(":"))
             } else if let Some(expr) = child.cast() {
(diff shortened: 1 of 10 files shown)


tests/src/unit.rs
```diff
@@ -96,6 +96,9 @@ pub fn collect_tests() -> Result<Vec<Trial>, Box<dyn Error>> {
 
 fn check_snapshot(path: &Path, width: usize) -> Result<(), Failed> {
     let (source, opt) = read_source_with_options(path)?;
+    if opt.use_simple_width && (width != 0 && width != 80) {
```
Enter-tainer, line 99: what does this change do?

QuadnucYard: for simple small documents, there's no need to test on so many widths. 0 and 80 are enough.

Enter-tainer: I don't like this idea. Testing them on more width doesn't hurt. One can also filter tests using cargo nextest cli

Enter-tainer (approved):

PR #282, opened 2025-05-11
feat: improve args layout with modified pretty.rs

QuadnucYard:


Cargo.toml
```diff
@@ -77,3 +77,4 @@ debug = true
 
 [patch.crates-io]
 # tinymist-world = { git = "https://github.com/Myriad-Dreamin/tinymist", rev = "84569fb98c4eeea65d91677c31d803718143fc10" }
+pretty = { git = "https://github.com/QuadnucYard/pretty.rs.git", rev = "106ee15a0ea3ca4a8f501ab8235d7bbc48dfa92f" }
```
Enter-tainer, line 80: consider publish it to crates io when we want to merge this. I remember if we have git deps, we will not be able to publish typstyle to crates.io

Issue #273, opened 2025-05-06
Math aligment: Overhang cells cause excessive gaps


MultisampledNight:
When a single alignment cell is split across multiple lines, unaffected lines cause expansion of their width.

### Current output

```typst
$
  forall n in NN :
  2 n              & = a
$
```

### Desired output

```typst
$
  forall n in NN :
  2 n &= a
$
```

- Rationale: The spacing between `2 n` and the equation sign made it hard to read.

(PR introducing `&` alignment: https://github.com/Enter-tainer/typstyle/pull/266)

QuadnucYard:
In your example, `forall n in NN :` is expected to participate in alignment.

Possible solutions:
- Make the expression taking up a source line not affect column width. That's what you desired.
- Source lines in a cell, if math expressions, align individually instead of as a whole. So it will become
  ```typ
  $
    forall n in NN :
                 2 n &= a
  $
  ```
(We should identify whether these solutions can behave badly in some other cases)

Enter-tainer:
FWIW, this it how it rendered in preview. @QuadnucYard could you please elaborate why this is intended? 

<img width="301" alt="Image" src="https://github.com/user-attachments/assets/aa68ee8e-20db-406f-9087-72eb62426fc2" />

QuadnucYard:
The math expression
```
  forall n in NN :
  2 n &= a
```
has two "cells" (2 columns):
```
forall n in NN :
2 n
```
and
```
= a
```
we compute the width of each column from the maximum line width in its cells.

Enter-tainer:
I'm not sure if i get it correctly: `forall n in NN :` and `2 n` should be in the same column. This is because there is no `\` after the first line. We can see this from the rendered image above

QuadnucYard:
It may be better to allow the last cell in a row "overflow".
In this way, there will not be so many spaces after `2 n`.

Enter-tainer:
fixed in #283. Now for multiline cells, we only used its last line to determine its width

PR #283, opened 2025-05-11
feat: calculate column width solely based on the last line of cells in math align

QuadnucYard:
related: #273


Enter-tainer (approved):
LGTM

PR #286, opened 2025-05-13
feat: always break after linebreak syntax and preserve linebreak in more cases in text-wrapping

### LINKED ISSUE #285: Markup linebreak (`\`) is ignored with `--wrap-text`
Before:

```typ
Text \
on the new line
```

After:

```typ
Text \ on the new line
```

I first faced this [here](https://forum.typst.app/t/how-do-i-automatically-change-colspan-of-data-cells-that-are-from-a-single-column/3993/2?u=andrew).

QuadnucYard:
Closes #285.

When text-wrapping is enabled:
- Always break after `\`.
- Preserve linebreak after:
  ```rs
  SyntaxKind::LineComment
      | SyntaxKind::BlockComment
      | SyntaxKind::Linebreak
      | SyntaxKind::Label
      | SyntaxKind::CodeBlock
      | SyntaxKind::ContentBlock
      | SyntaxKind::Conditional
      | SyntaxKind::WhileLoop
      | SyntaxKind::ForLoop
      | SyntaxKind::Contextual
  ```

Tell me if more cases should be covered.


crates/typstyle-core/src/pretty/markup.rs
```diff
@@ -279,11 +279,38 @@ impl<'a> PrettyPrinter<'a> {
...
+            is_block_elem(node) || matches!(node.kind(), SyntaxKind::LineComment)
+        }
+
+        fn prefer_break_after(node: &&SyntaxNode) -> bool {
+            is_block_equation(node) || matches!(node.kind(), SyntaxKind::Linebreak)
+        }
+
+        fn prefer_break_before(node: &&SyntaxNode) -> bool {
             is_block_equation(node)
         }
```
Enter-tainer, line 296: nice naming. Would it be possible if `cannot break before` and `should break after` happen at the same time?

Enter-tainer (approved):

crates/typstyle-core/src/pretty/markup.rs
QuadnucYard: no. in `should_break_after`, the node must be the last one in a line, and the break point is at the end.

crates/typstyle-core/src/pretty/markup.rs
```diff
@@ -279,11 +279,38 @@ impl<'a> PrettyPrinter<'a> {
     }
 
     fn convert_markup_body_reflow(&'a self, ctx: Context, repr: &MarkupRepr<'a>) -> ArenaDoc<'a> {
-        fn can_turn_exclusive(node: &&SyntaxNode) -> bool {
+        fn cannot_break_before(node: &&SyntaxNode) -> bool {
```
Enter-tainer, line 282: why `&&SyntaxNode`?

PR #298, opened 2025-05-21
feat: format tables in most cases

QuadnucYard:


QuadnucYard:
This prepares for table alignment.

For better visualization, it is better to put cells in a row on a single line, even if they are too long.
However, if any cell in a row is multiline, that is worse than putting each cell on an individual line.
In this case, break-suppressing will not help.

We may limit table alignment to cases where cells in the row are all single-line.

crates/typstyle-core/src/pretty/layout/table.rs
```diff
@@ -0,0 +1,103 @@
...
+pub struct TableCollector<'a> {
+    arena: &'a Arena<'a>,
+    /// When columns == 0, we will not reflow cells.
+    columns: usize,
+    row_docs: Vec<ArenaDoc<'a>>,
+    rows: Vec<Row<'a>>,
+}
+
+enum Row<'a> {
+    Cells(ArenaDoc<'a>, bool),
```
Enter-tainer, line 14: what does the bool mean? consider using struct enum item?

crates/typstyle-core/src/pretty/layout/table.rs
```diff
@@ -0,0 +1,103 @@
+use pretty::{Arena, DocAllocator};
+
+use crate::pretty::ArenaDoc;
+
+pub struct TableCollector<'a> {
+    arena: &'a Arena<'a>,
+    /// When columns == 0, we will not reflow cells.
+    columns: usize,
+    row_docs: Vec<ArenaDoc<'a>>,
+    rows: Vec<Row<'a>>,
```
Enter-tainer, line 10: do we need row_docs and rows at the same time?

crates/typstyle-core/src/pretty/layout/table.rs
```diff
@@ -0,0 +1,103 @@
...
+        if n == 1 && self.columns == 0 || n > 1 {
+            self.flush_cells();
+        }
+        if n > 1 {
+            self.disable_last_auto_break();
+            self.rows.push(Row::Linebreak);
+        }
+    }
+
+    fn flush_cells(&mut self) {
```
Enter-tainer, line 58: okay, so `row_docs` is `current_row_docs`, i guess?

Enter-tainer (approved):
overall LGTM

Enter-tainer (approved):

PR #302, opened 2025-05-23
feat: do not collapse consecutive spaces in markup by default

### LINKED ISSUE #296: Unwanted removal of spaces in Markup: breaking "Code only" & "Correctness"
I am trying to write a custom inline numbered list. Due to the lack of special grammar, I wanted to minimize the syntax needed. This is my version of the code (MRE):
```
#let opt(..content) = {
  let c = content.pos()
  let num = it => numbering("A. ", it)
  grid(
    columns: c.len() * (1fr,),
    ..c.enumerate().map(((i, v)) => num(i + 1) + v)
  )
}

#let opts(content) = {
  let arr = content.children.filter(it => "text" in it.fields()).map(it => it.text)
  opt(..arr)
}

1. #opts[randomness   potential     masterpiece   success]
2. #opts[emergency    maintenance   review        shade]
//instead of: 2. #opt[emergency][maintenance][review][shade]
```
However, after running the formatter, "extra" spaces are removed, and my markup became:
```
1. #opts[randomness potential masterpiece success]
2. #opts[emergency maintenance review shade]
```

This changes the semantics from a `sequence` (of `text`s interspersed by `space`s) to a consecutive `text`, it not only BREAKS the indention of the source code, but also cause my custom function to STOP working. 

I wonder why space removal is also applied to markups(The contents in brackets). I thought a formatter should have focused on "code only" as stated in the readme. I searched for relevant info in the doc, but it seems that only math mode can survive the formatting. 

I am really confused about this current behavior. Is it intended? If it is, I hope for an new option that allows spacing in markup to be left untouched (and hopefully this option doesn't violate the "opinionated" goal). If not so, is there any workaround that I can do now? Or is there a better way to implement such custom function(e.g. `#opts`) so that my markup survive the formatting?

Extra screenshots are provided for your convenience as follows:

Image
The "sampled values" is a sequence

Image
Now it is not.

QuadnucYard:
resolves #296.

now markup spaces between items will not collapse to a single one, unless enabling that feature.

### DIFF
--- a/README.md
+++ b/README.md
@@ -53,10 +53,10 @@ Options:
   -V, --version  Print version
 
 Format Configuration:
-  -l, --line-width <LINE_WIDTH>      The column width of the output [default: 80] [aliases: column] [short aliases: c]
-  -t, --indent-width <INDENT_WIDTH>  Spaces per level of indentation in the output [default: 2] [aliases: tab-width]
-      --no-reorder-import-items      Whether not to reorder import items
-      --wrap-text                    Whether to wrap texts in the markup
+  -l, --line-width <LINE_WIDTH>      Maximum width of each line [default: 80] [aliases: column] [short aliases: c]
+  -t, --indent-width <INDENT_WIDTH>  Number of spaces per indentation level [default: 2] [aliases: tab-width]
+      --no-reorder-import-items      Disable alphabetical reordering of import items
+      --wrap-text                    Wrap text in markup to fit within the line width. Implies `--collapse-spaces`
 
 Debug Options:
   -a, --ast         Print the AST of the input file
--- a/crates/typstyle-core/src/config.rs
+++ b/crates/typstyle-core/src/config.rs
@@ -2,16 +2,18 @@
 #[derive(Debug, Clone, PartialEq, Eq, Hash)]
 #[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
 pub struct Config {
-    /// Number of spaces per tab
+    /// Number of spaces to use for each indentation level.
     pub tab_spaces: usize,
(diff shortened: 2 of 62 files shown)


README.md
```diff
@@ -53,10 +53,11 @@ Options:
   -V, --version  Print version
 
 Format Configuration:
-  -l, --line-width <LINE_WIDTH>      The column width of the output [default: 80] [aliases: column] [short aliases: c]
-  -t, --indent-width <INDENT_WIDTH>  Spaces per level of indentation in the output [default: 2] [aliases: tab-width]
-      --no-reorder-import-items      Whether not to reorder import items
-      --wrap-text                    Whether to wrap texts in the markup
+  -l, --line-width <LINE_WIDTH>      Maximum width of each line [default: 80] [aliases: column] [short aliases: c]
+  -t, --indent-width <INDENT_WIDTH>  Number of spaces per indentation level [default: 2] [aliases: tab-width]
+      --collapse-spaces              Collapse consecutive whitespace in markup into a single space
```
Enter-tainer, line 58: do we need a config for this? i think we can remove the config and enable it by default

QuadnucYard: is just the same behavior as before when enabling it.

Enter-tainer: sorry, i mean we should always disable it

Enter-tainer: sorry, i mean we should always disable it

QuadnucYard: i'd like to keep it as an option for "breaking" rewrite (may break repr).
we may support more "breaking" text rewrite in the future, such as add spaces between cjk and latin chars, and prose wrap between sentences.

Enter-tainer (approved):

PR #320, opened 2025-06-13
feat: improve fitting and performance of compact layout

QuadnucYard:
Related: #282. This is a workaround, where we partially rewrite the dependent `pretty.rs`.

In the new union layout, whether it is fitted only depends on the first line. Therefore, following long lines will no longer prevent the compact layout. This is reasonable because they will not even fit in more expanded layouts. Moreover, the performance is guaranteed.

Now the only arg is no longer forced folded. A compact layout can handle them in most cases.


Enter-tainer (approved):

QuadnucYard:
superseded by #353

PR #322, opened 2025-06-14
feat: do not add extra space to empty code block

QuadnucYard:
align with the behavior of formatters of other languages


Enter-tainer (approved):

Enter-tainer:
Does rustfmt add space for empty code block?

QuadnucYard:
> Does rustfmt add space for empty code block?

Neither rustfmt nor biome adds that extra space. It seems that it's also true for most formatters.

PR #328, opened 2025-06-23
test: add llm-generated testcases

QuadnucYard:
Some can be also used as playground samples.

waited for #396 to make new tests pass


QuadnucYard:
blocked by #362 

in this case:
```typst
#{
data.filter(item =>
    criteria.keys()
  )
}
```
the closure is surprisingly regarded as **multiline**, simply because the first space in args is after `keys()` and contains a linebreak.

QuadnucYard:
blocker:
```typst
text 1.
```
can reflows to 
```typst
text
1.
```
where `1.` becomes an enum item.

QuadnucYard:
waited for #362: there are many bad cases related to closures

PR #332, opened 2025-06-25
Update links in README.md

rmburg:
Previously, the readme contained many dead links, due to a change of the repo URL.
This PR modifies all links to point to the new repo URL, except for the link to the [benchmarks](https://enter-tainer.github.io/typstyle-bench-results/dev/bench/) which is still in the `enter-tainer` namespace.

I noticed one more issue: the link in the CI badge at the top of the readme currently points to `...`, which does not exist. This should be fixed in another PR, but I can also include a fix in this one if so desired.


Enter-tainer (approved):
thanks!

PR #340, opened 2025-06-30
feat: do not use compact layout when args have multiline flavor

QuadnucYard:
(docs not updated)


QuadnucYard:
There is something wrong with compact layout, but i have no idea how it should be:
```typ
#{
  range(amount).map(i => init_function())
}
```
This can exceed width limit.

Enter-tainer (approved):
rules mentioned in pr description sounds good to me. will review snapshots later

Enter-tainer (approved):
snapshots look better than before

PR #359, opened 2025-07-10
feat: treat line comments as line suffixes

QuadnucYard:
WIP: wait for upstream feature stablized

Now line comments no longer take up any space.

Question: should it also apply to trailing block comments?

### DIFF
--- a/crates/typstyle-core/src/pretty/comment.rs
+++ b/crates/typstyle-core/src/pretty/comment.rs
@@ -13,10 +13,10 @@ enum CommentStyle {
     Bullet,
 }
 
-/// Convert either line comment or block comment.
+/// Convert either line comment or block comment. Line comments are converted as line suffixes.
 pub fn comment<'a>(arena: &'a Arena<'a>, node: &'a SyntaxNode) -> ArenaDoc<'a> {
     if node.kind() == SyntaxKind::LineComment {
-        line_comment(arena, node)
+        line_comment(arena, node).as_line_suffix()
     } else if node.kind() == SyntaxKind::BlockComment {
         block_comment(arena, node)
     } else {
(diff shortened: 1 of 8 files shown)


QuadnucYard:
> I wonder if we can reflow comments to next line. Overall I like this feature because in other languages I constantly see formatter breaking my code because it tries to squeeze comments and exprs into a single line. But on the other hand, making comments arbitrarily overflow the line width is not a good idea either 🤔

IMO it would be less descriptive if we reflow the comment to the next line. Additionally, users may comment out a suffix of the current line, and to reflow will break its structure.

Enter-tainer:
> > I wonder if we can reflow comments to next line. Overall I like this feature because in other languages I constantly see formatter breaking my code because it tries to squeeze comments and exprs into a single line. But on the other hand, making comments arbitrarily overflow the line width is not a good idea either 🤔
> 
> IMO it would be less descriptive if we reflow the comment to the next line. Additionally, users may comment out a suffix of the current line, and to reflow will break its structure.

I agree with the first point. How do you comment on the overflow problem?

QuadnucYard:
> > > I wonder if we can reflow comments to next line. Overall I like this feature because in other languages I constantly see formatter breaking my code because it tries to squeeze comments and exprs into a single line. But on the other hand, making comments arbitrarily overflow the line width is not a good idea either 🤔
> > 
> > 
> > IMO it would be less descriptive if we reflow the comment to the next line. Additionally, users may comment out a suffix of the current line, and to reflow will break its structure.
> 
> I agree with the first point. How do you comment on the overflow problem?

I would move the suffix comment to a new line before the item it attaches to, but this may not be everyone's choice. It is hard to judge whether this comment is suitable as a suffix by a machine.
After all, overflow is common in markup.

PR #362, opened 2025-07-11
feat: make flavor detection more intuitive

### LINKED ISSUE #356: Space preservation in markup
# The Problem

In markup, we may deliberately use spaces to give texts semantics. However, on the other hand, we may occasionally insert redundant spaces and want to tidy them up. To preserve or collapse or trim, that's the question.

Here is a tiny example to show how we treat spaces currently:

```typst
    leading spaces are trimmed

=   In Heading
-   In list item
+   In enum item

/  spaces in terms  :  are   trimmed


spaces   between   words   are preserved   without    `--collapse-space`

Spaces can turn to line breaks in the following cases:

#[  in content block  ]
#[but not in this one  ]

* in strong syntax *
* in 
multiline 
strong syntax *
_
  in emph syntax
_


#[- with blocky content inside]
#[$ "equation is also blocky" $]
#[= But heading is ok]
```

which with `-l=0` can be formatted to

```typst
leading spaces are trimmed

= In Heading
- In list item
+ In enum item

/ spaces in terms: are   trimmed


spaces   between   words   are preserved   without    `--collapse-space`

Spaces can turn to line breaks in the following cases:

#[
  in content block
]
#[but not in this one  ]

* in strong syntax *
*
in
multiline
strong syntax
*
_
in emph syntax
_


#[
  - with blocky content inside
]
#[$
    "equation is also blocky"
  $]
#[
  = But heading is ok
]
```

It is worth noting that, in the last case, it does not affect the layout to add extra spaces. But this indeed changes the representation, which is a matter of concern.

One tricky case is
```typst
#[- 1
   - 2
-   3]
```
As you can see, sticking the first list marker to the bracket `[` makes it confusing. So we format it as
```typst
#[
  - 1
    - 2
  - 3
]
```
But it also changes the representation.

## Special Rules

There are some existing options to change the default behavior.

### `--collapse-space`

Consecutive spaces are trimmed to one.

### `--wrap-text`

It implies `--collapsed`. Any space can turn into a line break.

# Discussion

- Should we preserve spaces as-is as much as possible? (which is conservative)
- How should the special rules interfere with the default behavior?

# Related

## Issues

- #296

## Pull Requests

- #286
- #302
- #307

QuadnucYard:
Previously, we checked if the first space contains any linebreak. But that would be counter-intuitive in view of spaces/linebreaks in children.
With this PR, a list-like syntax is regarded as multiline, if there is a space containing linebreak before its first child expr.

For example:
```typst
#arguments(red,stroke: blue,
 yellow)
#arguments(red, stroke: blue,
 green)
#arguments(stroke:
 blue, red)
```
Previously, the first was multiline flavored, but it differs from the second one only in a space.
Now none of them is multiline flavored.
To make it multiline flavored, you exactly need to insert a linebreak before the first arg.

For another,
```typst
#arguments(it => 
  it
)
```
the linebreak before `)` surprisingly makes it multiline-flavored.

Math/Markup flavor detection is not changed.
We need to resolve #356 for them first.

### DIFF
--- a/docs/pages/features/code.typ
+++ b/docs/pages/features/code.typ
@@ -6,7 +6,7 @@
 
 = Basic Layout
 
-Generally, there are some Typst syntaxes classified as "list-like":
+Typstyle classifies several Typst syntaxes as "list-like" structures:
 - Code block (```typc { a }```)
 - Array (```typc (1, 2, 3)```)
 - Dictionary (```typc (a: 1, b: 2)```)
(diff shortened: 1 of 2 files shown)


Enter-tainer:
@claude explain this PR to me in both Chinese and English. Mainly focus the main changes and potential problems. also tell me why this pr make sense

Enter-tainer:
pls also resolve the conflict

QuadnucYard:
> Hi, could you please update the pr description and include the latest algo for new flavor detection algorithm? I wonder if we should change docs files as well

Updated

Enter-tainer:
I wonder if we can minimize the behavior change for users. Would it make sense if we take a list as multiline if `old_method || new_methord` is true? Thus users will not see changes after upgrading to latest typstyle

QuadnucYard:
> I wonder if we can minimize the behavior change for users. Would it make sense if we take a list as multiline if `old_method || new_methord` is true? Thus users will not see changes after upgrading to latest typstyle

It has no effect on well-formatted code, since the two methods give the same results on well-formatted code.

docs/pages/features/code.typ
```diff
@@ -15,35 +15,49 @@ Generally, there are some Typst syntaxes classified as "list-like":
...
-- *Always-folded*: items are always put on one line regardless of width limit
-- *Never-folded*: items are spread across multiple lines
-- *Fitted*: items are folded if fitted the line
+- *Flat*: items are always put on one line regardless of width limit
+- *Expanded*: items are spread across multiple lines
 - *Compact*: (currently only applied to arguments) initial items are placed on the first line while the last combinable item spans multiple lines without extra indentation
 
-== Flavor Detection
+Typstyle choose the best layout with the following rule:
+- If it has multiline flavor, use *flat* only.
```
Enter-tainer, line 23: Do you mean expanded only?

Enter-tainer (approved):

PR #389, opened 2025-07-24
Add a nix flake

tingerrr:
See commit messages for more info.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,3 +3,6 @@
 
 /github-pages
 .claude
+
+# nix build output
+/result
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -10,6 +10,7 @@ edition = "2021"
 authors = ["Wenzhuo Liu"]
 description = "Beautiful and reliable typst code formatter"
 license = "Apache-2.0"
+readme = "README.md"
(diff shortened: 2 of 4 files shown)


Enter-tainer:
Hi, thank you for your contribution. However, I have some concerns regarding the maintainability of Nix flake files. Currently, none of active typstyle developers use Nix, which raises the possibility that these files might not always be functional. What do you think about this?

Besides, could you please explain more about what flake files are used for?

tingerrr:
> Currently, none of active typstyle developers use Nix, which raises the possibility that these files might not always be functional.

That's a reasonable concern. I think it'll mostly be fine if you just update the rust manifest version in the flake alongside the cargo manifest. I'd be willing to maintain the flake if it is added and no one else uses nix.

> Besides, could you please explain more about what flake files are used for?

Flakes are used by the nix package manager to install or run packages directly from git, but within the nix system (this is especially necessary when they need to link to system libraries). The lock file is as the name implies a file that pins the dependencies such that they should always work if no inputs change drastically.

Enter-tainer:
Thank you for explanation. I prefer not adding flake files at this momoent

PR #393, opened 2025-07-25
feat: use UTF-16 indices for range operations in wasm bindings

QuadnucYard:


crates/typstyle-wasm/src/lib.rs
```diff
@@ -51,26 +53,16 @@ pub fn format_ir(
 /// The result of formatting a range within content.
 #[wasm_bindgen(getter_with_clone)]
 pub struct FormatRangeResult {
-    /// Start byte offset (in UTF8) of the actual formatted range
+    /// Start UTF-16 index of the actual formatted range
```
Enter-tainer, line 56:
```suggestion
    /// Start UTF-16 code unit index of the actual formatted range
```

Enter-tainer (approved):

PR #396, opened 2025-07-26
feat: further preserve boundary spaces and improve text wrapping

### LINKED ISSUE #401: [Bug]: Unstable styling with wrap-text
### What happened?

Running typstyle twice is not stable:
```
❯ typstyle -i thesis.typ -l 100 -t 2 --wrap-text -v
```
Turns this:
```
To this neural network, we also add another convex term:
$
  [frac(abs(x_i)^k x_i, ell)]_i
$ <eqn:odd_power_term>
```
into this
```
To this neural network, we also add another convex term: $
  [frac(abs(x_i)^k x_i, ell)]_i
$ <eqn:odd_power_term>
```
and back, ad infinitum.

IMO, the $ should stay on the following line.

QuadnucYard:
related: #356. fixes #401
mainly resolves issues found from llm generated testcases.

- Without text-wrapping, spaces will not turn to linebreaks, unless they are next to blocky elements (list item, heading, etc).
- With text-wraping, always turn spaces around blocky equations and raws into linebreaks.

The current behavior of boundary handling (taking the left as an example):
- `[a`: -> `[a`
- `[   a`: -> `[   a` by default (preserved), -> `[ a` with collapse-spaces, `[ a` or `[(lf)a` with wrap-text
- `[(lf)a`: -> `[(lf)a` (preserved)
- `[ = a`: -> `[= a` or `[(lf)= a`

Fixed: `#[/*  */ 11 /*  */]` was formatted to `#[ /*  */ 11 /*  */ ]`, with unintended spaces prepended and appended.


Enter-tainer:
> Without text-wrapping, spaces will not turn to linebreaks, unless they are next to blocky elements (list item, heading, etc).


Does this effectively disables the folloing formatting?

```typst
if true [ 1 ] else [ 2 ] 
```

to

```
if true [
  1
] else [
  2
]
```

tests/fixtures/unit/code/snap/if-chain.typ-120.snap
```diff
@@ -46,8 +44,7 @@ input_file: tests/fixtures/unit/code/if-chain.typ
   true
   false
 } else if c < d [
-  false
-] else if 1 + 2 == 3 [*strong*] else if false { if 0 < 1 { true } } else {
+  false ] else if 1 + 2 == 3 [*strong*] else if false { if 0 < 1 { true } } else {
   a = 3
```
Enter-tainer, line 48: these changes don't look good

tests/fixtures/unit/code/snap/if-chain.typ-40.snap
```diff
@@ -8,15 +8,16 @@ input_file: tests/fixtures/unit/code/if-chain.typ
...
+  1 + 2 == 3
+) [*strong*] else { none }
 
 #if a < b { true } else if c < d [
-  false
-] else if 1 + 2 == 3 [*strong*] else {
+  false ] else if (
+  1 + 2 == 3
+) [*strong*] else {
   none
```
Enter-tainer, line 21: same here

tests/fixtures/unit/table/snap/large-cell.typ-40.snap
```diff
@@ -7,18 +7,14 @@ input_file: tests/fixtures/unit/table/large-cell.typ
...
-    [
-      #read(
+      ) ],
+    [ #read(
         "large-cell.typ",
         encoding: "utf8",
-      )
-    ],
+      ) ],
   ),
```
Enter-tainer, line 18: same

tests/fixtures/unit/code/snap/if-chain.typ-120.snap
QuadnucYard: that's intentionally keep spaces as is

Enter-tainer: how about the following examples?

Enter-tainer:
I'm not sure if I get it correctly. How will we "leave the decision to users"? I didn't see this pr adding a config item for this.(BTW I prefer not adding one).

From code, it looks like in this pr, we remove tight & symm rules and only depends on text wrapping. For users who already depends on tight & symm rules, this is a regression. Would it make sense if we keep tight & symm rules, but also take text wrapping into consideration?

QuadnucYard:
> I'm not sure if I get it correctly. How will we "leave the decision to users"? I didn't see this pr adding a config item for this.(BTW I prefer not adding one).

Users can change the boundary space to linebreaks if they find the layout bad, which is also respected by the formatter. 

> From code, it looks like in this pr, we remove tight & symm rules and only depends on text wrapping. For users who already depends on tight & symm rules, this is a regression. Would it make sense if we keep tight & symm rules, but also take text wrapping into consideration?

The boundary part of this PR makes formatting more conservative, so it will not change existing codebase.
When text wrapping is enabled, we do not preserve any space. Boundary spaces can turn into linebreaks, just as inline spaces.

QuadnucYard:
> If the tight rule is causing divergence, would it be possible if we remove tight rule and only keep the symm rule?

The symm rule targets this case: `[ content]`, `[content ]`. It becomes implicit as we choose *always* to preserve spaces.

The key point is whether `[ content ]` can become
```
[
  content
]
```
Considering #296, I want to preserve
```
[ aaaaa  bbbbbbb cccccccc   ]
[ dddd   eeee    ffffff     ]
```
where we manually align texts as DSL. In this case, we do not want a wide content block to be broken.

And I forgot one thing. With space collapse enabled, it should keep the original behavior.

Enter-tainer (approved):

NeilGirdhar:
Fixes #401 

(Adding comment so that it will auto-close the bug.)

PR #411, opened 2025-08-02
fix: never insert extra break at nil boundary for headings

QuadnucYard:
see https://github.com/typstyle-rs/typstyle/issues/356#issuecomment-3146173576


Enter-tainer:
pls update changelog.md, thx

PR #420, opened 2025-08-07
feat: experimental doc comment formatting

QuadnucYard:
## Summary

Doc comments (`/// comment`), although not admitted by the official, are already widely used in existing projects.
This PR introduces basic support for doc comment formatting. 

We create a sub-formatter for doc comments, which inherits the config of its parent.
- If `wrap_doc_comments` is set to `true`, the inner config also sets `wrap_text` to `true`.
- The inner `max_width` uses parental `doc_comment_width`.

Since we do not have a prefix element in the format engine, currently we have to print the contents of doc comments into strings, and then place them back to the document.

## Changes

- 
- 
- 

## Checklist

Before submitting, please ensure you've done the following:

- [ ] **Updated CHANGELOG.md**: Added your changes with examples to the changelog
- [ ] **Updated documentation**: Updated relevant docs, examples, or README
- [x] **Added tests**: Added tests for new features or bug fixes

### DIFF
--- a/crates/typstyle-core/src/config.rs
+++ b/crates/typstyle-core/src/config.rs
@@ -16,6 +16,13 @@ pub struct Config {
     /// When `true`, text in markup will be wrapped to fit within `max_width`.
     /// Implies `collapse_markup_spaces`.
     pub wrap_text: bool,
+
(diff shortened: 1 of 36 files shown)


QuadnucYard:
found a problem with text-wrapping:
the break-suppressing rule should be also applied

Issue #496, opened 2026-08-29
[Feature]: Preserve or configure line endings when formatting files


Gravifer:
### Description

Typstyle currently normalizes formatted output to LF line endings. As a result, a Typst file that is otherwise fully formatted but uses consistent CRLF line endings is reported by `typstyle --check` as requiring reformatting.

This is inconvenient for repositories that intentionally use CRLF through EditorConfig or Git checkout policy, especially on Windows.

### Minimal reproduction

Create an otherwise canonically formatted Typst file with CRLF line endings. 
Then, using Typstyle 0.15.1:

-----

```powershell
[IO.File]::WriteAllBytes(
  "example.typ",
  [Text.Encoding]::UTF8.GetBytes("#let value = 1`r`n")
)

typstyle --check example.typ
$LASTEXITCODE
```

<details>
<summary><i>(the same in other shells)</i></summary>

#### Git Bash

```bash
printf '#let value = 1\r\n' > example.typ

typstyle --check example.typ
echo $?
```

#### Command Prompt

```batch
>example.typ echo #let value = 1

typstyle --check example.typ
echo %ERRORLEVEL%
```
</details>

-----

The command exits with status 1. `typstyle --diff example.typ` reports a change even though the Typst formatting is already canonical and only the line endings differ.

Running `typstyle --inplace example.typ` rewrites the file with LF line endings.

### Current behavior

The formatter produces LF output, and check mode compares that output directly with the original input. Consequently:

- consistently formatted CRLF files fail `--check`;
- `--diff` may display the whole file as changed;
- `--inplace` changes the repository's line-ending convention;
- a repository using `end_of_line = crlf` cannot use `typstyle --check` directly.

### Expected behavior

For file inputs, Typstyle should ideally preserve the file's existing consistent line-ending style, or provide an explicit way to select it.

One possible interface would be:

```text
--line-ending auto | lf | crlf
```

Suggested semantics:

- `auto`: detect and preserve each input file's consistent line ending;
- `lf`: emit LF;
- `crlf`: emit CRLF;
- mixed line endings: either report an error or normalize according to an explicit option;
- stdin/stdout may continue to use LF by default.

In check mode, the formatter output would first be converted to the selected or detected line-ending style before comparison with the input.

The core formatter could continue using LF as its canonical internal representation; detection and restoration could remain a CLI file-I/O concern.

### Alternatives

[...]

Issue #434, opened 2025-09-28
[Feature]: Respect indenting from .editorconfig


tippfehlr:
### What would you like?

It would be nice if typstyle could get indent style and size from .editorconfig, or from the editor itself (I don’t know what would be possible).

### Why?

At the moment I use typstyle, and then format the indentation with nvim’s = separately. It would be great if typstyle did that automatically.

PR #206, opened 2024-12-25
fix: do not add additional end-of-file newline

ramdomPTM:
As discussed in #205, this PR fixes the printing and prevent typstyle from adding unintended additional end-of-file newlines.

PR #497, opened 2026-08-29
feat(cli): add explicit line-ending policies

Gravifer:
## Summary

Addresses #496.

This PR adds an explicit CLI choice for formatted-source line endings:

```text
--line-ending <lf|crlf-preserve>
```

The default remains `lf`, preserving the current behavior of `--check`,
`--diff`, `--inplace`, stdin, and stdout. Users can opt into conservative CRLF
preservation without changing `typstyle-core` or the formatter library APIs.

## Changes

- Add `--line-ending` with two named policies:
  - `lf` leaves formatter output unchanged and is the default.
  - `crlf-preserve` restores CRLF only when the input contains at least one
    CRLF and no bare LF or bare CR; otherwise it leaves formatter output
    unchanged.
- Apply an explicitly selected policy to the candidate formatted source before
  file/stdin output, `--check`, `--diff`, or `--inplace` handling. Diff control
  lines retain their normal LF output.
- Recognize only ASCII CR and LF as physical line endings. The policy
  postprocessor never rewrites VT, FF, NEL, U+2028, or U+2029 as line-ending
  sequences.
- Keep policy handling in `format_one`, after formatting but before the final
  changed/unchanged comparison. `format_debug`, `typstyle-core`, and the
  formatter library APIs remain independent of CLI output policy.
- Restore CRLF across the complete formatted output, including multiline
  strings and raw blocks, only after the conservative input check succeeds.
- Add byte-level CLI regression tests, focused policy tests, README usage,
  CLI usage documentation, and a changelog entry.

## Checklist

Before submitting, please ensure you've done the following:

- [x] **Updated CHANGELOG.md**: Documented the new CLI option, default, and
  policy boundaries.
- [x] **Updated documentation**: Updated the README CLI help and CLI usage
  documentation with both line-ending policies.
- [x] **Added tests**: Added policy, CLI parsing, help, and byte-level
  integration coverage.

## Testing

The following commands pass locally on Windows:

```console
cargo test -p typstyle
cargo check --workspace
cargo clippy --workspace --all-targets --all-features -- -D warnings
cargo fmt --check --all
git diff --check
```

Coverage includes:

- default and explicit `lf` retaining the existing formatter output behavior;
- formatted and unformatted consistent-CRLF input under
  `crlf-preserve` for check, diff, inplace, stdin, and stdout;
- LF, bare CR, no-newline, mixed CRLF/LF, and mixed CRLF/bare-CR behavior;
- repeated formatting and inplace-then-check convergence;
- changed diff output with CRLF source lines;
- CRLF restoration inside multiline strings and raw blocks;
- non-ASCII separators and invalid CLI values.

## Review Focus

Feedback is especially welcome on the public CLI shape and policy boundary:

- Is keeping `lf` as the backward-compatible default appropriate?
- Is the enum-shaped `--line-ending` option preferable to a narrower boolean
  such as `--preserve-line-endings`?
- Should forced CRLF output and configuration-file integration remain separate
  follow-up work?

### Why the policies are explicit

Automatically preserving CRLF would change the existing meaning of
`typstyle --check`: projects that currently use it to enforce LF would begin
accepting consistent-CRLF files. Keeping `lf` as the default makes the change
backward-compatible while allowing CRLF behavior to be selected deliberately.

`crlf-preserve` is intentionally not named `auto`. It recognizes only the
specific case that can be restored conservatively: the input contains at least
one CRLF and no bare LF or bare CR. Inputs without a line ending retain the
formatter output rather than forcing CRLF.

### Known limitations

- Pure CR preservation and forced CRLF output are deferred.
- Mixed inputs deliberately retain formatter output; preserving structural
  CRLF independently from semantic token contents would require source mapping
  or deeper core/printer support.
- This does not add EditorConfig or formatter configuration-file support. It
  remains independent of the broader configuration discussion in #115 and the
  closed EditorConfig request #434.
- The default end-of-file newline and stdout behavior established around #174
  and #206 remains unchanged; non-LF output requires an explicit policy.

### Commit history

The branch currently retains the exploratory commits and the explicit revert
that led to these policies. History cleanup remains deferred until the behavior
and scope are agreed during review.


QuadnucYard:
This workaround is acceptable, as it doesn’t require a core API change. Anyway, we need to expose this option to users.

crates/typstyle/tests/test_line_endings.rs
QuadnucYard, line 1: What's this? Pls well organize your test files

Gravifer: oops! i guess i've left this half-baked since last time

I'll take a look

Gravifer: Moved all line-ending coverage into a dedicated black-box integration test suite, removed the production-source `#[path]` inclusion, and restored the existing test files to their upstream layout.

Gravifer:
Separately (and unrelated to the line-ending changes in this PR),
I noticed a few correctness edge cases in the recently merged `fill-sentence` mode from #500:

- a non-breaking space can be removed during sentence trimming;
- inserting a sentence break before adjacent `-`, `+`, or `=` text can change how Typst parses it;
- a sentence immediately followed by a link can receive two line breaks.

I put together a tentative fix with regression tests on [`codex/fix-fill-sentence-boundaries`](https://github.com/Gravifer/QuadnucYard.typstyle/tree/codex/fix-fill-sentence-boundaries); the isolated change is [`20aa7de`](https://github.com/Gravifer/QuadnucYard.typstyle/commit/20aa7de5).

I won't burden you with a PR yet; take your time ;)

PR #498, opened 2026-08-31
fix: per sentence wrapping with common multi-dot abbreviations

tedil:
I noticed that abbreviations like `e.g.,` or `(i.e.,` were triggering false-positive sentence breaks with `--wrap-text=sentence`.

This happened because `convert_text_sentence_per_line` was passing the entire text slice into `is_common_abbreviation` instead of checking the (relevant) last word with `ends_with_common_abbreviation`, and surrounding punctuation (like trailing commas or parentheses) prevented exact matches.

So this PR basically now checks for `ends_with_common_abbreviation` (and strips boundary punctuation before checking abbreviations).
While at it, added a generic `is_multidot_acronym` helper for patterns like `z.B.` or `U.S.`; let me know if that is either too broad/heuristic-y or too specific ;) as well as some common abbreviations (`cf.`, `viz.`, `ibid.`, `ca.`, `eq.`, `ref.`) as well as a few extra boundary punctuation markers suggested by AI (feel free to ditch those if they feel redundant/unnecessary).

Also added a test case for the scenario mentioned above.


crates/typstyle-core/src/pretty/markup/wrapping.rs
```diff
@@ -272,31 +272,47 @@ fn sentence_ends_with_punctuation(text: &str) -> bool {
...
+                | '›'
+                | '‹'
+                | '*'
+                | '_'
+                | '`'
+                | '$'
+                | ','
+                | ';'
+                | ':'
+        )
```
QuadnucYard, line 311: I'd like to know if there is a better way to write such membership testing. It spans too many lines.

QuadnucYard:
Found two behavior regressions:

- **[P2] Numeric values are treated as acronyms**. `is_alphanumeric()` accepts `1.2.` and `1.2.3.`, so `The value is 1.2. Next sentence.` no longer breaks before `Next`. Restrict acronym components to letters.

- **[P2] Abbreviation trimming changes sentence-end detection**. Reusing `is_boundary_punctuation` strips commas, semicolons, and colons when detecting sentence endings. For example, ``He said `stop.,` then left.`` now gains a newline before `then`. Keep abbreviation normalization separate from sentence-closer trimming.

Confirmed both against the base revision. All six sentence-mode tests pass, but the three new examples also produce their expected output on the base revision. Workspace unchanged.

(Reviewed by GPT6)

tedil:
> Found two behavior regressions:
> 
>     * **[P2] Numeric values are treated as acronyms**. `is_alphanumeric()` accepts `1.2.` and `1.2.3.`, so `The value is 1.2. Next sentence.` no longer breaks before `Next`. Restrict acronym components to letters.
> 
>     * **[P2] Abbreviation trimming changes sentence-end detection**. Reusing `is_boundary_punctuation` strips commas, semicolons, and colons when detecting sentence endings. For example, `` He said `stop.,` then left. `` now gains a newline before `then`. Keep abbreviation normalization separate from sentence-closer trimming.
> 
> 
> Confirmed both against the base revision. All six sentence-mode tests pass, but the three new examples also produce their expected output on the base revision. Workspace unchanged.
> 
> (Reviewed by GPT6)

Indeed, I also noticed that and am slightly confused because for the document I was working on the behaviour was somehow different but I cannot reproduce it yet (i.e., the respective sentences in isolation do not seem to trigger the issue).
I'll give an update once I manage to produce a minimal example.

QuadnucYard:
Sentence segmentation is inherently language-dependent, but in our setting the input text has no associated locale or language specification.

Unicode sentence-boundary algorithms such as UAX#29 can identify generic candidate boundaries from character properties, punctuation, and surrounding structure. However, many ambiguous cases cannot be resolved from Unicode properties alone. In particular, a period may represent a sentence terminator, an abbreviation marker, an initial, part of an acronym, a decimal number, a version, or another non-terminal construct.

For example:

```text
Dr. Smith arrived.
Fig. 3 shows the result.
A. Smith proposed this.
U.S. researchers reported ...
```

Determining that the periods after `Dr.`, `Fig.`, or the components of `U.S.` do not terminate sentences normally requires lexical or linguistic knowledge. Such knowledge varies by language and sometimes by domain: the same token or punctuation pattern may have different interpretations in different languages or technical contexts.

When the locale is unknown, there is therefore no generally correct abbreviation-aware sentence segmentation algorithm. A segmenter must trade off between two kinds of error:

* **False positive boundary:** splitting at punctuation that is actually part of an abbreviation or other non-terminal construct.
* **False negative boundary:** failing to split where a sentence actually ends.

The problem is especially difficult for short fragments, because automatic language identification may itself be unreliable or impossible. Attempting to infer the locale can therefore introduce unstable segmentation behavior rather than eliminate the ambiguity.

The practical problem is consequently not to recover linguistically perfect sentence boundaries, but to define a deterministic, language-agnostic segmentation policy with acceptable error characteristics. A reasonable approach is to use Unicode sentence boundaries as candidate boundaries and suppress those that can be identified safely from language-independent structural evidence, such as decimals, version numbers, initials, acronym-like forms, URLs, code-like tokens, or explicitly configured abbreviations. Cases requiring language-specific lexical knowledge must remain inherently ambiguous unless additional language or domain information is provided.

As such, we might need to detect language from text styles (if available), or accept new formatter options for explicit control. But a mix of languages makes it even harder.

TheJanzap:
Would close #501

TheJanzap:
I have also played around with this issue and found that by simply looking at the last word, the problematic inputs I tested all disappeared. I also ran the entire test suite against this patch without problems. See [my branch](https://github.com/TheJanzap/typstyle/commits/fix-501/) for more.

```patch
diff --git a/crates/typstyle-core/src/pretty/markup/wrapping.rs b/crates/typstyle-core/src/pretty/markup/wrapping.rs
index 096b017c..645d6f6d 100644
--- a/crates/typstyle-core/src/pretty/markup/wrapping.rs
+++ b/crates/typstyle-core/src/pretty/markup/wrapping.rs
@@ -227,7 +227,8 @@ fn convert_text_sentence_split<'a>(
                 };
             }
             first = false;
-            previous_was_abbreviation = is_common_abbreviation(sentence);
+            let last_word = sentence.split_ascii_whitespace().last().unwrap_or("");
+            previous_was_abbreviation = is_common_abbreviation(last_word);
             ended_sentence = source_ends_with_sentence(sentence);
         }
         start = end;
```

Issue #499, opened 2026-09-15
Feature request: a wrap mode that reflows sentences to the print width (between `fill` and `sentence`)


cdesaintguilhem:
#405 asked for one-sentence-per-line formatting so that prose isn't "randomly reflow[ed]" the way `--wrap-text=fill` reflows it, and #483 (resolving #405) added `wrap_mode = "sentence"` / `--wrap-text=sentence`.

In practice, `sentence` mode only changes *where hardline breaks are inserted* (at sentence boundaries) — it does not reflow text to the configured print width at all:

- **A sentence longer than the line width is left on a single, unbounded line.** For example:

  ```typst
  All human beings are born free and equal in dignity and rights and they are endowed with reason and conscience and should always act towards one another in a spirit of brotherhood.
  ```

  ```
  $ typstyle --wrap-text=sentence input.typ
  All human beings are born free and equal in dignity and rights and they are endowed with reason and conscience and should always act towards one another in a spirit of brotherhood.
  ```

  That's 180+ columns on one line — `fill` would have wrapped it at 80.

- **A sentence that a user manually wrapped across multiple lines is left as-is instead of being reflowed.** For example:

  ```typst
  All human beings are born free
  and equal in dignity and rights.
  ```

  ```
  $ typstyle --wrap-text=sentence input.typ
  All human beings are born free
  and equal in dignity and rights.
  ```

  `fill` joins this back into `All human beings are born free and equal in dignity and rights.` on one line, but `sentence` leaves the manual line break untouched.

This makes `sentence` mode only predictable for text where every sentence already fits on one line and is written as a single source line, which limits its usefulness for real prose.

**Feature request:** add a mode that is a middle ground between `fill` and `sentence`: reflow each sentence's content to the print width, joining/breaking words as needed regardless of how the source happened to be wrapped *and* insert hardline at sentence boundaries — essentially "`fill`, but with a forced hardline between sentences" rather than "`sentence`, with no reflow at all."

PR #500, opened 2026-09-15
feat: add fill-sentence wrap mode

cdesaintguilhem:
Addresses #499.

Adds `wrap_mode = "fill-sentence"` / `--wrap-text=fill-sentence`, a middle ground between `fill` and `sentence`: like `sentence`, it forces a hardline at each sentence boundary, but like `fill`, it reflows each sentence's words to the print width instead of leaving it as a single unbounded line or preserving the source's manual line breaks.

- `bbcfbe8` unifies the `fill` and `sentence` markup-reflow code paths (`convert_markup_body_wrapped`) so both share the same width-aware line-filling logic, parameterized by whether sentence boundaries force a hardline.
- `7ca7fe3` adds the `FillSentence` wrap mode using that shared path.
- Remaining commits add snapshot/CLI test coverage and update the CLI help and docs for the new mode.


QuadnucYard:
LGTM

PR #502, opened 2026-09-21
feat: Ensure single final newline

humtta:
## Summary

This PR ensures that formatted files end with exactly one newline.

## Changes

- Normalize the formatter output to remove trailing newlines and append one final newline.
- Add unit tests covering inputs with no trailing newline and with multiple trailing newlines.
- Document the feature in CHANGELOG.

## Testing

Compiled and ran the binary on files with no trailing newlines and with multiple trailing newlines. Also ran the unit tests; all of them passed successfully.


QuadnucYard:
This is intended behavior.

humtta:
I might be missing something, but is there any reason why this is the expected behavior? Formatters usually normalize extra (or missing) trailing newlines.

QuadnucYard:
> I might be missing something, but is there any reason why this is the expected behavior? Formatters usually normalize extra (or missing) trailing newlines.

This is markup, not code in the general sense. Blank lines can be intentional. As long as we preserve newlines in the body, the ending should be no exception.

Of course, we could have an option to control the behavior. But by default, we tend to preserve. You can still turn on the trimming option in VSCode if you like, as an alternative.

humtta:
That's fine. If you think it's reasonable, I can implement a `--trim-final-newlines` flag in another PR.

